PR #5, opened 2020-03-20
Eflumerf/initial daq process

eflumerf:
This pull request is not intended to be a merge request; it is simply to set up a place for discussion of the code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,3 +1,10 @@
+cmake_minimum_required(VERSION 3.12)
+project(app-framework)
+
+set(CMAKE_CXX_STANDARD 17)
+
+include_directories(${CMAKE_SOURCE_DIR}/app-framework/include)
+include_directories(${CMAKE_SOURCE_DIR}/app-framework-base/include)
 
 add_subdirectory(apps)
 add_subdirectory(doc)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,8 @@
+
+add_subdirectory(app)
+add_subdirectory(doc)
+add_subdirectory(include)
+add_subdirectory(src)
+add_subdirectory(test)
+add_subdirectory(tools)
+add_subdirectory(unittest)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,8 @@
 
-add_subdirectory(app)
+add_subdirectory(apps)
 add_subdirectory(doc)
 add_subdirectory(include)
+add_subdirectory(scripts)
 add_subdirectory(src)
 add_subdirectory(test)
-add_subdirectory(tools)
 add_subdirectory(unittest)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,9 @@ project(app-framework)
 
 set(CMAKE_CXX_STANDARD 17)
 
+find_package(TRACE 3.15.0 REQUIRED)
+
+
 include_directories(${CMAKE_SOURCE_DIR}/app-framework/include)
 include_directories(${CMAKE_SOURCE_DIR}/app-framework-base/include)
 
--- a/apps/simple_test_app.cc
+++ b/apps/simple_test_app.cc
@@ -14,7 +14,7 @@ std::unique_ptr<CommandFacility> CommandFacility::handle_ = std::unique_ptr<Comm
 
 class simple_test_app_ModuleList : public ModuleList {
     // Inherited via ModuleList
-    virtual void ConstructGraph(BufferMap& buffer_map, UserModuleMap& user_module_map, TransitionOrderMap& transition_order_map) override {
+    virtual void ConstructGraph(BufferMap& buffer_map, UserModuleMap& user_module_map, CommandOrderMap& command_order_map) override {
         user_module_map["debugLogger"].reset(new DebugLoggingUserModule());
     }
 };
--- a/apps/simple_test_app.cc
+++ b/apps/simple_test_app.cc
@@ -3,27 +3,26 @@
 #include "app-framework/QueryResponseCommandFacility.hh"
 
 namespace appframework {
-	std::unique_ptr<CommandFacility> CommandFacility::handle_ = std::unique_ptr<CommandFacility>(new QueryResponseCommandFacility());
+std::unique_ptr<CommandFacility> CommandFacility::handle_ = std::unique_ptr<CommandFacility>(new QueryResponseCommandFacility());
 
-	class simple_test_app_ModuleList : public ModuleList {
-		// Inherited via ModuleList
-		virtual void ConstructGraph(BufferMap& buffer_map, UserModuleMap& user_module_map, TransitionOrderMap& transition_order_map) override
-		{
-			user_module_map["debugLogger"].reset( new DebugLoggingUserModule());
-		}
-	};
-}
+class simple_test_app_ModuleList : public ModuleList {
+    // Inherited via ModuleList
+    virtual void ConstructGraph(BufferMap& buffer_map, UserModuleMap& user_module_map, TransitionOrderMap& transition_order_map) override {
+        user_module_map["debugLogger"].reset(new DebugLoggingUserModule());
+    }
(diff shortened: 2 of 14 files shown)


apps/simple_test_app.cc
```diff
@@ -0,0 +1,28 @@
+#include "app-framework/DAQProcess.hh"
+#include "app-framework/DebugLoggingUserModule.hh"
+#include "app-framework/QueryResponseCommandFacility.hh"
+
+namespace appframework {
+std::unique_ptr<CommandFacility> CommandFacility::handle_ = std::unique_ptr<CommandFacility>(new QueryResponseCommandFacility());
```
eflumerf: This should be replaced by module loading code ASAP

include/app-framework/DAQProcess.hh
```diff
@@ -0,0 +1,77 @@
...
+     * @brief Start the CommandFacility listener
+     * @return Return code from listener
+     *
+     * This function should call the loaded CommandFacility::listen method, which should block for the
+     * duration of the DAQ Application, calling execute_transition as necessary.
+     */
+    int listen();
+
+   private:
+    StateMachineState myState_;
```
mroda88, line 69: The state machine is defined as the *base* level. Maybe we don't need to standardise this especially if the state is CCM business.

src/DAQProcess.cc
```diff
@@ -0,0 +1,47 @@
...
+void DAQProcess::register_modules(std::unique_ptr<ModuleList> const& ml) {
+    ml->ConstructGraph(bufferMap_, userModuleMap_, transitionOrderMap_);
+    myState_ = StateMachineState::ModulesRegistered;
+}
+
+void DAQProcess::execute_transition(TransitionName cmd) {
+    if (transitionOrderMap_.count(cmd)) {
+        for (auto& moduleName : transitionOrderMap_[cmd]) {
+            if (userModuleMap_.count(moduleName)) {
+                userModuleMap_[moduleName]->execute_transition(cmd);
```
mroda88, line 36: Both the execute_transision() calls should be inside a try/catch loop. Question we have to answer is: what do we do if we catch something? Probably at this point, only dying.

include/app-framework/DAQProcess.hh
mroda88: We might need to make this status available to the outside. In this case we might need to make this an atomic variable and add a reader method of this quantity.

src/QueryResponseCommandFacility.cc
```diff
@@ -0,0 +1,42 @@
...
+int QueryResponseCommandFacility::listen(DAQProcess* process) {
+    try {
+        bool keepGoing = true;
+        while (keepGoing) {
+            std::cout << "Enter a command" << std::endl;
+            std::string comm;
+            std::getline(std::cin, comm);
+
+            if (comm == "quit") {
+                keepGoing = false;
```
mroda88, line 23: I'm fully aware that this is a place holder until we have a CCM interface. Yet there is a logical behaviour which I believe is dangerous. Having the interface to know the concept of the command "quit" so that it can let the process die is a bit dangerous: requires re implementation of the quit command in different places. 
I suggest this interface to be agnostic in terms of the command. The Command facility (and later the CCM) can simply ask the state to the DAQProcess and exit this loop once the state is something like "waiting to be killed". 
If we go toward this direction, we can give a sense to the concept of state at the level of the application framework and reduce the implementation of the CCM interface.

CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+include_directories(${CMAKE_SOURCE_DIR}/app-framework/include)
+include_directories(${CMAKE_SOURCE_DIR}/app-framework-base/include)
+
+add_subdirectory(apps)
+add_subdirectory(doc)
+add_subdirectory(include)
+add_subdirectory(scripts)
+add_subdirectory(src)
+add_subdirectory(test)
+add_subdirectory(unittest)
```
brettviren, line 15: Is it possible to:

* extract this boiler plate into a file that a package `#include`s in some way (instead of copy-pastes) 
* to not supply,eg `apps/` but to have `add_subdirectory()` not fail?

What's I'm hoping CMake can do is like what `waf` lets us.  A very minimal per-package build file.  Eg:

https://github.com/WireCell/wire-cell-toolkit/blob/master/img/wscript_build

It names itself and any dependencies and that's it.

apps/simple_test_app.cc
```diff
@@ -0,0 +1,35 @@
...
+    std::list<std::string> args;
+    for (int ii = 1; ii < argc; ++ii) {
+        args.push_back(std::string(argv[ii]));
+    }
+
+    appframework::DAQProcess theDAQProcess(args);
+
+    auto ml = std::unique_ptr<appframework::ModuleList>(new appframework::simple_test_app_ModuleList());
+    theDAQProcess.register_modules(ml);
+
```
brettviren, line 33: It would be good to package this even further.  It seems `ml` is wholly internal and nothing that the `main()` cares about.  Further factoring would allow multiple `DAQProcess` instances to run in separate threads of a given process.

One major benefit of the actor pattern is to have the flexibility to aggregate actors in processes or actors as processes.

apps/simple_test_app.cc
```diff
@@ -0,0 +1,35 @@
...
+};
+}  // namespace appframework
+
+int main(int argc, char* argv[]) {
+    std::list<std::string> args;
+    for (int ii = 1; ii < argc; ++ii) {
+        args.push_back(std::string(argv[ii]));
+    }
+
+    appframework::DAQProcess theDAQProcess(args);
```
brettviren, line 29: The `args` should be some kind of pre-parsed pre-validated configuration object.

include/app-framework/DAQProcess.hh
```diff
@@ -0,0 +1,77 @@
...
+
+    BufferMap bufferMap_;                    ///< String alias for each Buffer
```
brettviren, line 71: I don't understand the `BufferMap` enough to comment but it's clearly central.  This needs explanation.

include/app-framework/DAQProcess.hh
```diff
@@ -0,0 +1,77 @@
...
+    UserModuleMap userModuleMap_;            ///< String alias for each UserModule
```
brettviren, line 72: I don't understand the `UserModuleMap` enough to comment but it's clearly central. This needs explanation.

apps/simple_test_app.cc
bieryAtFnal: Could multiple DAQProcesses instances in a given process talk to each other (presumably via the same types of Buffers/Queues that UserModules within a DAQProcess can talk to each other)?  I suspect yes...

src/DAQProcess.cc
```diff
@@ -0,0 +1,45 @@
...
+
+void DAQProcess::register_modules(std::unique_ptr<ModuleList> const& ml) {
+    ml->ConstructGraph(bufferMap_, userModuleMap_, commandOrderMap_);
+}
+
+void DAQProcess::execute_command(std::string cmd) {
+    if (commandOrderMap_.count(cmd)) {
+        for (auto& moduleName : commandOrderMap_[cmd]) {
+            if (userModuleMap_.count(moduleName)) {
+                userModuleMap_[moduleName]->execute_command(cmd);
```
mroda88, line 34: This block needs two things  that might be done.
1) try - catch method
2) check the status of the future so that we can report a timeout or something 
Not sure we need them now.

src/DAQProcess.cc
```diff
@@ -0,0 +1,45 @@
...
+            }
+        }
+    } else {
+        for (auto const& um : userModuleMap_) {
```
mroda88, line 38: this default order of transition should disappear. It is the CommandOderMap that should provide a default. Otherwise logically we define things in different places. Not healthy.

eflumerf: The idea here was that the CommandOrderMap only defines required orderings, and the implementation should decide for anything that is not specified. However, there should definitely be a warning message printed if a command does not have any entry in the commandOrderMap_.

PR #6, opened 2020-04-13
Wketchum/initial buffer class

wesketchum:


bieryAtFnal:
As a test, I added the creation of a DequeBuffer<int> to simple_test_app.cc.  
In order to get that to compile, I needed to make a few changes to the buffer classes:
- push and pop both needed to be changed to return size_t instead of int in DequeBuffer and FollyDeque
- the (templated) method declarations in DequeBuffer.cc and FollyDeque.cc needed to be moved into the corresponding header files
- DequeBuffer<T>::pop() needed to have "fDeque.pop_front(obj)" changed to "fDeque.pop_front()"

include/app-framework/Buffers/DequeBuffer.hh
```diff
@@ -0,0 +1,39 @@
...
+  public:
+
+    DequeBuffer();
+
+    void Configure();
+    
+    size_t size() { return fDeque.size(); }
+    size_t capacity() { return fCapacity; }
+
+    int push(const T&); ///push one on, return new size if successful, -1 if not
```
mroda88, line 23: This interface is in contrast with the Buffer interface.

src/Buffers/DequeBuffer.cc
```diff
@@ -0,0 +1,45 @@
...
+  }
+  else{
+    fCapacity = fDeque.max_size();
+  }
+
+}
+
+template <class T>
+int appframework::DequeBuffer<T>::push(const T& obj){
+
```
mroda88, line 24: Do we need a thread safety mechanism here?

src/Buffers/DequeBuffer.cc
```diff
@@ -0,0 +1,45 @@
...
+    fDeque.push_back(obj);
+    return size();
+  }
+
+  return -1;
+
+}
+
+template <class T>
+int appframework::DequeBuffer<T>::pop(T& obj){
```
mroda88, line 35: Same here: are we happy with a non-thread safety mechanism?

include/app-framework/Buffers/DequeBuffer.hh
```diff
@@ -0,0 +1,45 @@
...
+    T   pop();          ///pop one off, return object
+
+  private:
+    
+    std::deque<T> fDeque;
+    size_t        fCapacity;
+
+    std::mutex fMutex;
+    int fRetryTime;
```
mroda88, line 37: shall we make it a duration? http://www.cplusplus.com/reference/chrono/duration/duration/

include/app-framework/Buffers/DequeBuffer.hh
```diff
@@ -0,0 +1,45 @@
...
+  template <class T>
+  class DequeBuffer : public BufferInput<T>, BufferOutput<T> {
+    
+  public:
+
+    DequeBuffer();
+
+    void Configure();
+    
+    size_t size() { return fDeque.size(); }
```
mroda88, line 22: This is dangerous because fDeque could be locked, so every access of fDeque should be locked. I suggest we have size counter which is atomic if order to deal with this.

src/Buffers/DequeBuffer.cc
```diff
@@ -0,0 +1,70 @@
...
+  fPushRetries = this->fPushTimeout/fRetryTime;
+  fPopRetries  = this->fPopTimout/fRetryTime;
+}
+
+template <class T>
+int appframework::DequeBuffer<T>::push(const T& obj){
+
+  int n_retries=0;
+  while(full()){
+    if(n_retries==fPushRetries){
```
alessandrothea, line 36: What is the rationale of handling the retries internally, rather than letting the module to decide how many retries are acceptable?

src/Buffers/DequeBuffer.cc
```diff
@@ -0,0 +1,70 @@
...
+  if(full()) throw std::runtime_error("Full after check? Another thread accessing?");
+  fDeque.push_back(obj);
+  return size();
+}
+
+template <class T>
+T appframework::DequeBuffer<T>::pop(){
+
+  int n_retries=0;
+  while(empty()){
```
mroda88, line 55: not thread safety here. But the problem is in the size function, not here

brettviren:
The comment on adding sleeps makes me think it would be useful to add a speed test analogous to this one:

https://github.com/brettviren/zio/blob/master/test/test_speed.cpp#L63

Or for that matter any tests! :)

philiprodrigues:
Shouldn't we be using an existing implementation of a concurrent queue, instead of trying to write our own? Then it's already tested and debugged

src/Buffers/DequeBuffer.cc
eflumerf: Both push and pop should have a timeout parameter to allow the UserModule pushing/popping an opportunity to go and do something else, give up, or throw an execption

eflumerf: In this case the retries are a consequence of the timeout configuration parameter defined in Buffer{Input,Output}. This is a specific implementation detail, different Buffer implementations may have their own ways of handling timeouts in push/pop.

It is of course up to UserModules to determine what to do when a push/pop times out.

src/Buffers/DequeBuffer.cc
eflumerf: Resolved in Marco's branch.

PR #9, opened 2020-04-22
Feature/style guide conformance

jcfreeman2:
Focus is on the DequeBuffer:

-Get it to pass the linters
-A major change has been to move the (template) function definitions into the DequeBuffer header, and (at least for now) get rid of the DequeBuffer code which was in a source file


include/app-framework/Buffers/DequeBuffer.hh
```diff
@@ -38,6 +48,75 @@ namespace appframework{
     size_t fPopRetries;
   };
 
+
+template <class T>
```
mroda88, line 52: This is exactly why I asked asked about the extension of templated file we want to use, when we had this discussion  ages ago. Would it be possible to put these implementation in a file called DequeuBuffer.icc  that is included at the end of the header?

jcfreeman2: If people agree that this a good idea, I personally have no problem with it. From my perspective, the positive is that you don't have tons of code right after the class declaration that a class's users (not developers) don't need to worry about, the negative is that developers would need to jump back and forth between the two files.

jcfreeman2:
The modification of the buffer code (code which derives from app-framework-base's BufferI.hh, in this case, DequeBuffer.hh) to conform to the style guide has been considerably updated since yesterday. A couple of points:

1) This code builds against app-framework-base's feature branch of the same name, specifically, commit faf18924c9ae7bb0356a2deafd8e4bf0ba09a65c

2) As discussed with Marco, whether to put the implementation of DequeBuffer in a separate *.icc file is to be decided

3) The copy and move operations of DequeBuffer are now defined - by explicitly deleting them. Why this was done is explained in the comments.

jcfreeman2:
Today's work involved:

-Getting the remaining files in conformance with the C++ Style Guide
-Getting the code up-to-date with the alterations I made to the handle() function interfaces, as described in today's comment on app-framework-base's feature/style_guide_conformance branch

jcfreeman2:
One final commit: DAQProcess:register_modules now takes just a reference to a ModuleList instance, rather than a unique_ptr to a ModuleList instance.

jcfreeman2:
At this morning's task force meeting we signed on to Marco's idea of moving the implementation of a template class's functions into an includable *.icc file, so that's been done for DequeBuffer. It was agreed that all other topics discussed related to the actual semantics of the code rather than their style would be dealt with on other branches. I consider this branch ready for review.

PR #10, opened 2020-04-23
DAQprocess refinement with future

mroda88:
This pull requests implement the simplest pattern that uses the future returned from a UserModule during a transition. So far this was unused. Once configuration will be more elaborate these patterns can be complicated depending on configuration or requirements from the UserModule implementations.

### DIFF
--- a/src/DAQProcess.cc
+++ b/src/DAQProcess.cc
@@ -16,6 +16,7 @@
 #define TRACE_NAME "DAQProcess"
 
 #include <unordered_set>
+#include <vector>
 
 namespace appframework {
 ServiceManager* ServiceManager::handle_ = nullptr;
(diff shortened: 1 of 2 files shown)


mroda88:
I added catches for different stages of the command propagation. The pull request is ready for review. This is a first exploration of having modules returning futures.

src/DAQProcess.cc
```diff
@@ -30,18 +31,26 @@ DAQProcess::DAQProcess(std::list<std::string> args) {
...
     if (commandOrderMap_.count(cmd)) {
         for (auto& moduleName : commandOrderMap_[cmd]) {
             if (userModuleMap_.count(moduleName)) {
-                userModuleMap_[moduleName]->execute_command(cmd);
-                user_module_list.erase(moduleName);
+	        try { 
+	            cmd_results.push_back( std::make_pair( moduleName, userModuleMap_[moduleName]->execute_command(cmd) ) ) ; 
+	        }
+	        catch ( ... ) {
+		  TLOG(TLVL_WARNING) << moduleName << " throw an exception while receiving command " << cmd ;
```
brettviren, line 50: This is a thorny problem, but what is any external entity supposed to do with this info?

Presumably this log message makes it's way to CCM (eventually) and CCM reacts or ignores it.

For this to have a chance of being usefully processed I think at least the following info needs to be added

* an identifier for the exception type that was caught
* some identifier of the DAQ process as a whole, both its type (program name) and it's unique ID.

mroda88: Hi Brett, yes. At the moment we don't have specific suggestions from the CCM. This is our expected behaviour. 

I agree with both your suggestions. Specifically.
> an identifier for the exception type that was caught

This is in the pipeline. Alessandro has the exception based code mentioned by Giovanna and once we are having a look at it, we will start implementing the dedicated specifications. Until then I'll prefer to keep this item on hold.

>some identifier of the DAQ process as a whole, both its type (program name) and it's unique ID. 

This is in fact very much desirable as it is one of the requirements from CCM we discussed at the CERN workshop. I'll try to see if I can code something simply that can be easily added to the DAQProcess interface.

eflumerf:
Rejected due to App Framework v1 Discussions.

PR #11, opened 2020-05-01
Buffer timeout interfaces

mroda88:


mroda88:
this pull request adapt preexisting code to the new buffer interfaces of the base. Simplification of the present buffer will be implemented later.

PR #13, opened 2020-05-08
Eflumerf/fake data user modules

eflumerf:
This pull request adds a FakeDataProducer and FakeDataConsumer, as well as a FanOutUserModule. These modules are loaded into a simple test application which shows vectors of integers moving through the application.


jcfreeman2:
Still OK, Eric's commit is just part of us getting everything onto develop.

PR #17, opened 2020-05-21
Queue(Source|Sink) implementation with ZeroMQ

philiprodrigues:
This draft pull request introduces appframework::ZeroMQueue, a QueueSink/QueueSource implementation which uses ZeroMQ. The implementation works by passing a pointer to the queue item as the content of the ZeroMQ message. This means it's limited to communication between modules in the same process, but there are no extra memory copies. In tests, I get about 800k msgs/second.

The implementation uses ZeroMQ PUSH/PULL sockets, which means that it's effectively an MPMC queue, with a fair-queueing implementation on input, and round-robin on output. If strict SPSC behaviour is desired, ZeroMQ PAIR sockets can be used.

At this point, I'm just looking to socialize the idea of using ZeroMQ in this way, as an approach that gives us _both_ lightweight intra-process messaging and a straightforward path to flexible inter-process communication. Substantive comments on how to improve the interface or performance are welcome, but it's probably too early for code-style comments.

Some changes that remain to be made:
* Need to implement `can_push()`, which will be similar to `can_pop()`
* I'm using the cppzmq C++ bindings for zeromq, which are header-only, so I just copied them in (`zmq.hpp` and `zmq_addon.hpp`). This should probably be moved to its own ups package
* Something is needed to allow PUB/SUB (ie fan-out with items copied to all subscribers). The most natural way to do this is to add serialization, probably via encoder/decoder functions passed to the `QueueSink/Source` ctors. This of course will have the beneficial side-effect of making inter-process communication possible too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,9 +5,11 @@ set(CMAKE_CXX_STANDARD 17)
 
 find_package(TRACE 3.15.0 REQUIRED)
 
-
 include_directories(${CMAKE_SOURCE_DIR}/app-framework/include)
 include_directories(${CMAKE_SOURCE_DIR}/app-framework-base/include)
+
+include_directories($ENV{ZEROMQ_INC})
+find_library(LIBZMQ zmq $ENV{ZEROMQ_LIB})
 
 add_subdirectory(apps)
 add_subdirectory(doc)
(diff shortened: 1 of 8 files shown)


philiprodrigues:
I implemented `can_push()` and set a send timeout. Also changed the way receive timeouts are done, which leads to a factor of >2 speedup, so we get something like 1.7m msg/s in the `zeromqueue_check` test

PR #20, opened 2020-06-01
Remove Issue header files

mroda88:


mroda88:
based on app-framework-base pull request https://github.com/DUNE-DAQ/app-framework-base/pull/27

PR #22, opened 2020-06-02
Ron/ers trace

ron003:
ers+trace functionality with consistent streamer interface
Issue #23


ron003:
I think I should (or need to) make a new pull request.

PR #24, opened 2020-06-02
QueueRegistry fix: missing return

alessandrothea:
Thanks to @philiprodrigues for spotting a silly mistake in `QueueRegistry::get_queue`.
When requesting the an existing queue, the lack of a return statement resulted in the queue to be re-created. Fixed.


alessandrothea:
Cheers!

PR #29, opened 2020-06-05
Small refinements

mroda88:
a few refinements to comply with previous issues


include/app-framework/QueueRegistry.hh
```diff
@@ -18,8 +18,22 @@
 #include <memory>
 #include <string>
 
+
+/**
```
eflumerf, line 22: Doxygen really doesn't like ERS_DECLARE_ISSUE; I found that I had to put them at the end of the header file so that Doxygen would parse the rest properly.

mroda88: In this case it cannot be: it is called in a function that is implemented inline. We can define the function in the cc if we prefer.

alessandrothea: This seems bizarre. ERS clearly has doxygen-style comments in it...

PR #40, opened 2020-06-16
Old interface folly

mroda88:
This is the implementation of the folly queue with the `T pop()` interface.


alessandrothea:
There seem to be a bunch of commits outside the scope of this PR which I thought had already been merged in develop, right?

mroda88:
They were and they are still there. But to restore the old interface there was a bit of git black magic to do and github sees in this way. The number of changes is limited to the scope.

alessandrothea:
Ok, I wonder if that was really necessary, but fine as long as the result is OK.

More to the point, the workaround to adapt pop(T&) to T pop() involves a new. Doesn’t that defy the purpose of using a high performance queue which is supposed to do memory management behind the scenes?

mroda88:
No. High performance for queue it's not related to memory allocation. It's related to memory access that is done in a thread safe way and with small latency: that's what an efficient queue is. 
This is usually achieved allocating all the memory of the queue at once at its construction time, which is why queues also have a fixed size and they are not extendable like the STL containers. In fact having a queue passing memory only via pointers makes the memory allocated by the queue smaller. 
The `new` operation is not an issue of this case, it is a requirements of whenever a piece of memory cannot be allocated: if the memory to be transported is not copy-able the new is a requirement to be used in the pop( T & ) case as well, you just use it on the pop side of the queue instead of the push side. Also, note that the all the constructors are called with move semantic and therefore they are not copies.

alessandrothea:
How is new not an issue? In order to achieve performance both pop and push must be fast, or low latency as you say. Memory allocation via new on **each** push is a practice to avoid because it pays the price of  heap allocation.

mroda88:
It's not an issue because it would happen anyway on the pop side. Also speed is guaranteed by the fact that no copies are made. If they were this code would not even compile. 
I did run the FollyQueue unittest in both cases: I see a difference of 94 microseconds in times of execution and no differences in the sanity checks times. So the performances seem similar to me.

alessandrothea:
Just to be clear, my issue is not where the `new` is, but that there's a `new` at all.
The folly queue implementation goes to a good length to only do placement new and this wrapper is basically voiding that effort.

The 94 us difference is between what and what? I don't see any statistically relevant test in the unittest of the sort that is usually used for benchmarks (e.g. long loops)

jcfreeman2:
Marco's approach (using a move constructor so a unique_ptr can have an address for something originally provided as an rvalue) doesn't strike me as particularly objectionable, although instead of using
```
std::move( std::unique_ptr<T>( new T (std::move(t) ) )) 
```
as the argument to push, I'd have expressed it as 
```
std::make_unique<T>( std::move(t) )
```
I suppose the question is how much focus we want to put on optimizing for speed before the v1 release.

alessandrothea:
Well, given that the folly queue is being introduced for performance, providing it via wrapper that degrades its performance is quite pointless. We might not have put black on white, but it kind of goes without saying that the `T pop()` implementation must match the `pop(T&)` performance, otherwise it's a no-go.

To make sure my worries are well founded and that I haven't misunderstood many posts on the subject, I've put together a straw man example to compare the performance of placement new vs new

https://github.com/DUNE-DAQ/appfwk/blob/thea/placement_new/test/placement_test.cxx

It tries to mimic the native behaviour of the folly queue by using placement new to set the content of a buffer (a pre-allocated vector of `Stuff`) and compares it with filling a vector of unique pointers instantiating a new pointer each time.

This is what I get, a O(100) difference.
```
Placement new: elapsed time : 0.00130188s
New:           elapsed time : 0.0182803s
```
In principle the test should be repeated with the folly queue, comparing the un-wrapped and wrapped performance, but for me this is sufficient to say, unless there are alternative ideas for `T pop()`, that `pop(T&)` is the choice for v1.

brettviren:
@alessandrothea your "placement new" test measurement gets ~2MHz of object creation.  That is a ballpark rate which I expect.  Something is dreadfully wrong to get only 60kHz for just object creation.  

I get a similar ~2MHz rate when I measure both push and pop and using simple copy with `std::deque` in:

 https://github.com/brettviren/zio/blob/master/test/test_speed.cpp  

In that test, the equivalent ZeroMQ "push/pop" provides ~1MHz for inproc.  This provides far more benefit than does `std::queue` so the 2x slowdown is compensated. 

We should stop wasting time reinventing square wheels when perfectly round ones exist.

jcfreeman2:
We can't really say we got a 30x speedup (!) since this is simply not an apples-to-apples comparison. The thing about the placement new vs. make_unique test Alessandro performed is that the Entity is actually having to get memory allocated on the heap for each push in his make_unique test. This is *not* what's happening when Marco's push is called - the object Marco's pushing is *already* on the heap, it's simply having ownership switched over to the queue via a move. 

Having said this, I like the benchmarking approach, and it'll be easy to write up a quick script to use queue_IO_check to test queue performance.

jcfreeman2:
To follow up on the benchmarking I mentioned in an earlier comment and which was discussed at a meeting this morning: 

On the OldInterfaceFolly branch I added FollyQueuePopT.hpp, which basically performs a simple pass-by-reference wrap of the folly library implementation's dequeuing functionality, i.e.
```
void pop( value_type& t, const duration_type& dur) {
    if ( !fQueue.try_dequeue_for( t, dur) ) {
      ...code which throws an exception...
}
```
I updated queue_IO_check so it could test this pass-by-reference wrapper queue (or both of them, rather- we have a wrapper around folly::DSPSCQueue and around folly::DMPMCQueue). I also wrote a script which calls queue_IO_check under multiple scenarios:

One thread pushing
One thread popping 
One thread pushing while another thread pops
Two threads pushing while two threads pop

and compared the performance of the different queues. The queues in FollyQueuePopT.hpp were roughly 20% faster when it came to pushing and 30-40% faster when it came to popping, as compared to the implementation described earlier on this page which uses unique_ptrs. There may be ways to speed that implementation up, but it's of course a question of priorities over the next  couple of days.

PR #44, opened 2020-06-18
Adding command registration

alessandrothea:
Adds to `DAQModule` the interfaces to enable derived classes to register methods as commands.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,7 @@ add_executable(queue_IO_check test/queue_IO_check.cxx)
 target_link_libraries(queue_IO_check ${DAQ_LIBRARIES_UNIVERSAL_EXE} )
 
 
+add_library(appfwk_DummyModule_duneDAQModule test/DummyModule.cpp)
 add_library(appfwk_DebugLoggingDAQModule_duneDAQModule test/DebugLoggingDAQModule.cpp)
 add_library(appfwk_FakeDataConsumerDAQModule_duneDAQModule test/FakeDataConsumerDAQModule.cpp)
 add_library(appfwk_FakeDataProducerDAQModule_duneDAQModule test/FakeDataProducerDAQModule.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,9 @@ point_build_to( test )
 
 add_executable(queue_IO_check test/queue_IO_check.cxx)
 target_link_libraries(queue_IO_check ${DAQ_LIBRARIES_UNIVERSAL_EXE} )
+add_executable(dummy_test_app test/dummy_test_app.cxx)
+target_link_libraries(dummy_test_app appfwk appfwk_DummyModule_duneDAQModule pthread ers  ${CETLIB} ${CETLIB_EXCEPT} ${Boost_PROGRAM_OPTIONS_LIBRARY})
+
 
 
 add_library(appfwk_DummyModule_duneDAQModule test/DummyModule.cpp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ configure_file(scripts/setupForRunning.sh.in scripts/setupForRunning.sh @ONLY)
 ##############################################################################
 point_build_to( src )
 
-add_library(appfwk SHARED src/QueueRegistry.cpp src/DAQProcess.cpp)
+add_library(appfwk SHARED src/QueueRegistry.cpp src/DAQProcess.cpp src/DAQModule.cpp)
 
 set( APPFWK_LIBRARIES ${DAQ_LIBRARIES_PACKAGE} )
 list(REMOVE_ITEM APPFWK_LIBRARIES appfwk)
@@ -74,7 +74,7 @@ point_build_to( unittest )
 add_unit_test(ThreadHelper_test)
 add_unit_test(DAQSink_DAQSource_test)
 add_unit_test(StdDeQueue_test)
-add_unit_test(FanOutDAQModule_test)
+# add_unit_test(FanOutDAQModule_test)
 
 
 
--- a/include/appfwk/DAQModule.hpp
+++ b/include/appfwk/DAQModule.hpp
@@ -29,6 +29,8 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <functional>
+
 
 #ifndef EXTERN_C_FUNC_DECLARE_START
 #define EXTERN_C_FUNC_DECLARE_START                                                                                    \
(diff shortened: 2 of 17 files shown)


CMakeLists.txt
```diff
@@ -99,6 +99,7 @@ add_executable(queue_IO_check test/queue_IO_check.cxx)
 target_link_libraries(queue_IO_check pthread ${Boost_PROGRAM_OPTIONS_LIBRARY} )
 
 
+add_library(appfwk_DummyModule_duneDAQModule test/DummyModule.cpp)
```
mroda88, line 102: we don't need a new test module to implement this idea

include/appfwk/DAQModule.hpp
```diff
@@ -99,10 +156,34 @@ class DAQModule : public NamedObject
    *  Non-accepted commands or failure should return an ERS exception
    * indicating this result.
    */
-  virtual void execute_command(const std::string& cmd, const std::vector<std::string>& args) = 0;
+  void execute_command(const std::string& name, const std::vector<std::string>& args = {});
```
mroda88, line 159: I recommend that execute_commands stays virtual to allow optimisations or specific behaviour. 
If we are changing this interface we might as well change the argument to a json.

include/appfwk/DAQModule.hpp
```diff
@@ -99,10 +156,34 @@ class DAQModule : public NamedObject
...
+
+  std::vector<std::string> get_commands() const;
+
+  bool has_command(const std::string name) const;
```
mroda88, line 163: for the same reason this also should go to virtual

mroda88: maybe let's add a reference to be consistent everywhere

include/appfwk/DAQModule.hpp
```diff
@@ -99,10 +156,34 @@ class DAQModule : public NamedObject
...
+    if (!done) {
+      // Throw here
+      throw CommandRegistrationFailed(ERS_HERE, name);
+    }
+  }
+
   json configuration_; ///< JSON configuration for the DAQModule
+
+private:
+  using CommandMap_t = std::map<std::string, std::function<void(const std::vector<std::string> &)>>;
```
mroda88, line 185: I suggest a better name: CommandFunctionMap

src/DAQModule.cpp
```diff
@@ -0,0 +1,36 @@
...
+      return;
+    }
+
+    throw UnknownCommand(ERS_HERE, name);
+}
+
+std::vector<std::string> 
+DAQModule::get_commands() const {
+    std::vector<std::string> cmds(commands_.size());
+    for ( const auto &[key, value] : commands_)
```
mroda88, line 26: In case the commads_ 's type is changed, I sussgest
```C+++
for ( const auto & it : command_ ) 
  cmds.push_back( it -> first ) 
```

include/appfwk/DAQModule.hpp
```diff
@@ -99,10 +156,34 @@ class DAQModule : public NamedObject
...
+  bool has_command(const std::string name) const;
 
 protected:
+  /**
+   * @brief Registers a mdoule command under the name `cmd`.
+   * Returns whether the command was inserted (false meaning that command `cmd` already exists)
+   */
+  template <typename Child>
+  void
+  register_command(const std::string &name, void (Child::*f)(const std::vector<std::string>&)) {
```
mroda88, line 172: This is a complex function: let's leave the implementation in the hxx and let's not change the order of the file. class definition first, issues last. No need to change the whole file only to have an inline function.

CMakeLists.txt
alessandrothea: It can be removed, I don't mind. But it may come back.
A do-nothing dumb-as-a-rock module tends to be useful for very basic unit tests.

include/appfwk/DAQModule.hpp
alessandrothea: This was done on purpose to avoid `DAQModule` bypass command registration.

include/appfwk/DAQModule.hpp
alessandrothea: Same answer as above. This is meant not to be virtual.

src/DAQModule.cpp
alessandrothea: How would that make a difference? Value's type is inferred and its' not used anyway.

include/appfwk/DAQModule.hpp
mroda88: Why? I'm not saying it has to be encouraged, but still not providing freedom for flexibility is a dangerous thing in a framework. The usage case for an overload will be that depending on the state of DAQ module the same command  can go to different functions.

src/DAQModule.cpp
mroda88: It's not a problem of the type. It's just that if the type changes the loop can be left unchanged.

include/appfwk/DAQModule.hpp
alessandrothea: Fair enough, let me do that.
Anyway, we need to sort out why doxygen and ers don't get along. I'll open an issue ™️

src/DAQModule.cpp
alessandrothea: The loop can be left unchanged only if `it` is an iterator exposing `->first`, which practically means a map iterator under any realistic scenario. Under the same scenario, the structured binding remains valid and it is more readable.

include/appfwk/DAQModule.hpp
alessandrothea: In all practical scenarios, because the additional freedom adds little in terms of flexibility while increasing the risk for users to bypass command creation. 
Allowing the same command having a different behaviour depending on the state, why wouldn't one handle such case in the command method?

mroda88: simple code modularity. You write the simplest function you can and if it's necessary you combine them as necessary.

include/appfwk/detail/DAQModule.hxx
```diff
@@ -0,0 +1,16 @@
+namespace dunedaq::appfwk {
+
+template <typename Child>
+void
+DAQModule::register_command(const std::string &name, void (Child::*f)(const std::vector<std::string>&)) {
+  using namespace std::placeholders;
+
+  bool done = commands_.emplace(name, std::bind(f, static_cast<Child*>(this), _1)).second;
```
mroda88, line 8: This `static_cast` is a problem in cases where the DAQModule has its own hierarchy. Since this call bypasses the access to the object vtable, if the methods that are called via this map are defined `virtual` we only call the function defined when register_command is invoked. It's a bit against OOP to be honest. Solutions might be possible but I have to think about it

eflumerf:
Resolves #55

include/appfwk/detail/DAQModule.hxx
alessandrothea: Not really. The cast here is necessary to bind a method (`Child::f`) to `DAQModule` which doesn't have it.
Further overriding of the same method in `Child` derived classes should be picked up correctly

https://stackoverflow.com/questions/35787752/how-to-store-a-virtual-member-function-in-stdfunction


https://stackoverflow.com/questions/35787752/how-to-store-a-virtual-member-function-in-stdfunction

mroda88: Yeah, it is the *should* that is not exactly making me happy. I had a look at the bind documentation and it's not very specific. I managed to find one other example on stackoverflow about this but it has not been confirmed by other users:  https://stackoverflow.com/questions/47986978/virtual-functions-and-stdfunction. 
My understanding is that `std::bind` is implemented in the a way so that the `Child` specification disappears from the call. That still opens the question on how to do the opposite: calling the base implementation, but this is not very important nor as problematic as the previous case. 
In general I tend not to base my code on something I read on stack overflow: users tend to do all sort of mess just to have the code running, but here we have to support some else's code as well as it's a framework. I honestly would like to hear someone else's opinion as well. My worry is that we are creating something that can pose limits in the future without gaining much. Let me elaborate on this. If a system like this could check that a module has implemented all the commands that are necessary (specified from a list of some sort) I would immediately see the benefit as we would have a compile time error if something is missing. As it is right now it's just an elaborate `switch case` with the limitations listed above.

include/appfwk/DAQModule.hpp
alessandrothea: Let's get a couple of things straight about the requirements on the framework, which still seem unclear.
The goal of the framework is to provide as much flexibility as possible, within the robustness and predictability requirements.
When it comes to module commands (stateless actions, let's not mix things up with FSM transitions), this translates into the module 
* having the list of commands defined by the end of initialisation
* the behaviour of each command predictable at all times
* the list of commands accessible by the caller (`DAQProcess` in our case)
* a user API that **enforces** within reason the previous requirements to be met.

A command registry is a typical pattern to achieve this (see `XDAQ` and others). 
Code modularity shall be encourage as much as possible within these boundaries.
So,
* mutable command behaviour is actually **not** desirable
* mutable behaviour of `execute_command` is not desirable
* state-aware commands are an option, they are not in scope for v1 and they will be equally regulated.
* there is still a lot of rope for users to hang themselves, if they have a deep desire to, for instance with hand-crafted ifs. If they do, they will be shown the wrongness of their ways.

include/appfwk/detail/DAQModule.hxx
alessandrothea: Right, in reverse order

I hope to have clarified that the limitations described above are the desired behaviour. Many examples of users overestimating their C++ abilities are available to back this up.
BTW, this is just the beginning. Further structures will be applied as the code evolves to point users in the right direction.

So, a quick test is a great idea. It's actually pretty quick, so I'm surprised you haven't taken the chance first.
It turns out that stack-overflow is right and the only reason why it didn't work in the first place was my naive use of `static_cast`. After replacing it with `dynamic_cast` the virtualness is preserved.

Example being pushed in the next few minutes, module github quirks.

include/appfwk/detail/DAQModule.hxx
```diff
@@ -0,0 +1,16 @@
+namespace dunedaq::appfwk {
+
+template <typename Child>
+void
```
eflumerf, line 4: Do we need the template? If we move the bind to the register_command call in the DAQModule implementation, then this can just use a std::functional argument, as in ThreadHelper.

alessandrothea: Alas yes, otherwise the compiler categorically refuses to assign the bound function to the `std::function` object.
This is the first thing I tried actually, and costed me a couple of hours of sleep...

Issue #54, opened 2020-06-18
Replace TRACE and std::exception


eflumerf:
Info, Warning, and Error messages should be handled through ERS. Likewise, our code should not throw std::exceptions but appropriate ERS Issues.

alessandrothea:
And debug.

PR #58, opened 2020-06-18
Replace exception with ers issues

eflumerf:
This branch resolves part of #54, namely replacing std::runtime_error with ers Issues.


eflumerf:
This pull request now completely resolves #54, and is ready for review.

PR #79, opened 2020-06-22
Include a "name" parameter to GeneralDAQModuleIssue

eflumerf:
I also changed the message text of GeneralDAQModuleIssue to something a bit more helpful. I still would like to know from someone more experienced with ERS if there is a way to print the base class message first...


alessandrothea:
Looks good. While we are at it, isn’t Generic in GenericDAQModuleIssue redundant?

Issue #76, opened 2020-06-22
Race condition in FanOutDAQModule_test


philiprodrigues:
From 10 runs of `FanOutDAQModule_test`:

```Running 4 test cases...
/home/rodrigues/dune/daq/daq-app-framework/src/old-build-method/appfwk/unittest/FanOutDAQModule_test.cxx(100): fatal error: in "FanOutDAQModule_test/NonCopyableTypeTest": critical check outputbuf1.can_pop() == true has failed [false != true]

*** 1 failure is detected in the test module "FanOutDAQModule_test"
Running 4 test cases...

*** No errors detected
Running 4 test cases...

*** No errors detected
Running 4 test cases...

*** No errors detected
Running 4 test cases...

*** No errors detected
Running 4 test cases...
/home/rodrigues/dune/daq/daq-app-framework/src/old-build-method/appfwk/unittest/FanOutDAQModule_test.cxx(100): fatal error: in "FanOutDAQModule_test/NonCopyableTypeTest": critical check outputbuf1.can_pop() == true has failed [false != true]

*** 1 failure is detected in the test module "FanOutDAQModule_test"
Running 4 test cases...

*** No errors detected
Running 4 test cases...

*** No errors detected
Running 4 test cases...

*** No errors detected
Running 4 test cases...

*** No errors detected```

The fraction of successes increases with the length of the sleep in the test

eflumerf:
I can run this test reliably with sleep lengths down to 2 ms. My build virtual machine has 16 cores and 64 GB of RAM, and I am the only user. What are the stats of the machine you are running on?

philiprodrigues:
The machine is a powerful one (Xeon(R) Gold 5115 CPU, 40 cores, 192 GB RAM), but well-used (47 users, load average 3). It seems generally not very responsive for me in interactive use. I'll try tomorrow with some other machines.

philiprodrigues:
I get the same race condition behaviour on dunegpvm11. Do you see the same thing there?

PR #81, opened 2020-06-23
Fix FanOutDAQModule_test

eflumerf:
This branch fixes Issue #76 by removing the race condition between the testing thread and the FanOutDAQModule working thread. I also reduced the default wait time in FanOutDAQModule when it is polling for input from 1s to 10,000 us.


eflumerf:
When I ran the updated test in my environment (where it was working reliably beforehand), it reported a delay of 1000 ms between the pushes to the input buffer and when the output buffers were ready. After changing the default wait time in FanOutDAQModule to 10 ms, the test reported that the processing took 10 ms, as expected.

PR #82, opened 2020-06-23
WIP: Build appfwk with "Modern CMake" conventions

philiprodrigues:
This WIP pull request modifies the appfwk build to follow "Modern CMake" conventions as described, for example, in https://cliutils.gitlab.io/modern-cmake/ . The idea is that by making appfwk follow commonly-used conventions, we make things easier to understand/use, and buy ourselves more flexibility later on (eg, when incorporating appfwk in a new build system).

This also allows us to meet the requirement that users of appfwk only have to build their own package, with minimal build instructions that look like:
1. Setup appfwk via package-management system (ups, spack)
2. Build your module as normal for a cmake project

"Modern CMake" style has some nice features:
* Each package installs a "config" file that allows it to be found with `find_package()` (provided `$CMAKE_PREFIX_PATH` is set appropriately by the package management system). The config file also finds the package's dependencies, so the user of the package doesn't need to
* Along with the "config" file, packages can make some of their targets available to dependent packages by "exporting" them. These targets carry dependency information along with the compile definitions required to compile/link against the target.

With the code in this branch, a minimal user package CMakeLists.txt would look something like:

```cmake
cmake_minimum_required(VERSION 3.12)
project(framework-user-test VERSION 1.0.0)

set(BUILD_SHARED_LIBS ON)
set(CMAKE_CXX_STANDARD 17)

find_package(appfwk REQUIRED)

# Build an executable. Only need to link against direct dependencies (eg, appfwk::appfwk).
# cmake takes care of transitive dependencies
add_executable(        framework-user-test-app framework-user-test-app.cpp)
target_link_libraries( framework-user-test-app appfwk::appfwk)

# Build a library (eg a DAQModule)
add_library(           my-test-module MyTestModule.cpp)
target_link_libraries( my-test-module appfwk::appfwk)

# Get ${CMAKE_INSTALL_*DIR} variables used below. Or could just specify them manually
include(GNUInstallDirs)

install(
    TARGETS framework-user-test-app my-test-module
    LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}"
    ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
    RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
    INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
)
```

Some points for discussion:
* I moved the `DAQ.cmake` and `Findcetlib.cmake` modules from `daq-buildtools` into `appfwk` and makes them visible to dependent packages. If we envisage a case where a package that doesn't depend on appfwk will want these modules, they can be moved back to daq-buildtools and said package can depend on daq-buildtools. (With the current cmake-packages branch, the only reason for daq-buildtools to exist is to give `quick-start.sh` somewhere to live).
* I only install `daq_application` and the `appfwk` library. We could consider also installing the executables from the `test/` directory
* I vendored `nlohmann/json.hpp` since it's just one header


brettviren:
+1

One comment:

> I vendored nlohmann/json.hpp since it's just one header

Just a heads up: this will eventually become problematic once appfmk depends on libraries which also depend on and expose `json.hpp` via library headers.  We will then need some method to assure the same version of `json.hpp` is used for both.

For right now, I think vendoring is reasonable as it gives convenience.  This issue kicks in only when this library dependency is formed as it will be when the CCM support library begins to be used by the appfmk.

philiprodrigues:
> Just a heads up: this will eventually become problematic once appfmk depends on libraries which also depend on and expose `json.hpp` via library headers. We will then need some method to assure the same version of `json.hpp` is used for both.

Good point. I've created a ups product of nlohmann::json in daq-externals which we can use post-v1

brettviren:
@philiprodrigues does your UPS product place the header at:

```
<prefix>/include/nlohmann/json.hpp
```
So that we will do

``` c++
#include "nlohmann/json.hpp"
```

This is proper. I only ask because I just now realized in past usage (PTMP) I wrongly placed it at `include/json.hpp` and did `#include "json.hpp` and I want to make sure my mistake isn't being followed.  I'll leave PTMP as-is for now but will follow the proper pattern going forward.

philiprodrigues:
@brettviren just checked, and it does the right thing for include path

philiprodrigues:
superseded by #103

PR #84, opened 2020-06-23
Convert pop() method to throw an exception rather than return a boolean

bieryAtFnal:
This pull request is to ask for approval to merge the changes on the biery/Issue73_pop_throws_exception branch into the develop branch.

The changes on that branch are ones needed to convert the Queue pop method from returning a boolean to throwing an exception when no element is available to pop within the specified timeout.

### DIFF
--- a/include/appfwk/DAQSource.hpp
+++ b/include/appfwk/DAQSource.hpp
@@ -38,7 +38,7 @@ public:
   using duration_type = std::chrono::milliseconds;
 
   explicit DAQSource(const std::string& name);
-  bool pop(T&, const duration_type& timeout = duration_type::zero());
+  void pop(T&, const duration_type& timeout = duration_type::zero());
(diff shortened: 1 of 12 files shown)


bieryAtFnal:
It would be great if any/all reviewers of this pull request would verify that the unit tests still work as expected after these changes.

eflumerf:
One question: Should the tests do a BOOST_TEST_REQUIRE(false, "Exception thrown") kind of message in their catch(){} blocks instead of simply ignoring the exception?

bieryAtFnal:
That sounds fine to me.  
What I tried to do was keep the existing behavior the same.  If the existing code ignored the boolean return value and there was a subsequent check that would have noticed that the pop() failed, I felt OK with simply ignoring the exception and using the following code to report the error. 
But, more information is good.

Issue #56, opened 2020-06-18
Bring codebase in line with the DUNE C++ Style Guide


jcfreeman2:
After some discussion, about a month ago it was decided that in the interests of speed and convenience it would no longer be required that code on a feature branch conform to the best-practices guidelines described in the C++ Style Guide (https://github.com/DUNE-DAQ/styleguide/blob/develop/dune-daq-cppguide.md) for it to be approved for merging. Instead, we'd do this after all the merging of feature branches into develop was complete. 

This issue should be addressed as the second-to-final one before declaring v1 complete (clang-formatting should be the final one).

jcfreeman2:
The v1.0 release is largely in compliance with the style guide; deviations tend to be of the unavoidable sort (e.g., using macros to deal with plugins, etc.)

PR #87, opened 2020-06-23
Eflumerf/lint tools compliance

eflumerf:
Run dune-cpp-style-check.sh and clang-format on appfwk source.


eflumerf:
Note that this is only the "easy" part of Issue #56, the code will still need to be inspected manually for other style guide violations.

PR #89, opened 2020-06-24
JCF: bring codebase significantly closer into compliance with the DUN…

jcfreeman2:
…E C++ Style Guide

The header mostly says it all. These are changes which can't (at
currently) get picked up by the linter. Here are some of changes made:

* Error checking
I made an effort to ensure that every function which could throw an exception is wrapped in a try-catch, and every
function which returns a status variable has that variable checked. As I made these changes under a time
constraint, where I was unable to do this, I added TODOs.

* const-ification
Many variables and functions which were logically const but not declared as such are now const.
GraphConstructor::ConstructGraph, CommandFacility::listen, etc.

* Removal of unnecessary interface
E.g., replaced non-owning pointers with references when the address of an object isn't of interest

* Removal of raw owning pointers
...and replacement with smart pointers

* Formatting of TODOs
I retroactively added names, dates and emails to TODOs which were missing these

### DIFF
--- a/src/QueueRegistry.cpp
+++ b/src/QueueRegistry.cpp
@@ -11,21 +11,18 @@
 #include "appfwk/QueueRegistry.hpp"
 
 #include <map>
+#include <memory>
 #include <string>
 
 namespace dunedaq::appfwk {
 
-QueueRegistry* QueueRegistry::me_ = nullptr;
-
-QueueRegistry::QueueRegistry() {}
-
-QueueRegistry::~QueueRegistry() {}
+  std::unique_ptr<QueueRegistry> QueueRegistry::me_ = nullptr;
 
 QueueRegistry&
 QueueRegistry::get()
 {
   if (!me_) {
-    me_ = new QueueRegistry();
+    me_.reset( new QueueRegistry() );
   }
   return *me_;
 }
--- a/test/FakeDataConsumerDAQModule.cpp
+++ b/test/FakeDataConsumerDAQModule.cpp
@@ -28,9 +28,6 @@ namespace dunedaq::appfwk {
 FakeDataConsumerDAQModule::FakeDataConsumerDAQModule(const std::string& name)
   : DAQModule(name)
   , thread_(std::bind(&FakeDataConsumerDAQModule::do_work, this, std::placeholders::_1))
-  , nIntsPerVector_(999)
-  , starting_int_(-999)
-  , ending_int_(-999)
   , queueTimeout_(100)
   , inputQueue_(nullptr)
 {
--- a/test/FakeDataConsumerDAQModule.cpp
+++ b/test/FakeDataConsumerDAQModule.cpp
@@ -50,6 +50,7 @@ FakeDataConsumerDAQModule::do_configure(const std::vector<std::string>& /*args*/
   nIntsPerVector_ = get_config().value<int>("nIntsPerVector", 10);
   starting_int_ = get_config().value<int>("starting_int", -4);
   ending_int_ = get_config().value<int>("ending_int", 14);
+  queueTimeout_ = std::chrono::milliseconds(get_config().value<int>("queue_timeout_ms", 100));
 }
 
 void
--- a/test/FakeDataConsumerDAQModule.cpp
+++ b/test/FakeDataConsumerDAQModule.cpp
@@ -28,6 +28,9 @@ namespace dunedaq::appfwk {
 FakeDataConsumerDAQModule::FakeDataConsumerDAQModule(const std::string& name)
   : DAQModule(name)
   , thread_(std::bind(&FakeDataConsumerDAQModule::do_work, this, std::placeholders::_1))
+  , nIntsPerVector_(999)
+  , starting_int_(-999)
+  , ending_int_(-999)
   , queueTimeout_(100)
   , inputQueue_(nullptr)
 {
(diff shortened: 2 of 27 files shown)


test/queue_IO_check.cxx
```diff
@@ -52,12 +52,12 @@ auto timeout = std::chrono::milliseconds(100); ///< Queue's timeout
  */
 std::unique_ptr<dunedaq::appfwk::Queue<int>> queue = nullptr;
 
-constexpr int nelements = 100; ///< Number of elements to push to the Queue (total)
+int nelements = 10000000; ///< Number of elements to push to the Queue (total)
```
eflumerf, line 55: Is there a reason for this to be so high? This executable takes ~30 seconds to run now. Maybe 1,000,000?

src/QueueRegistry.cpp
```diff
@@ -11,21 +11,18 @@
...
-
-QueueRegistry::~QueueRegistry() {}
+  std::unique_ptr<QueueRegistry> QueueRegistry::me_ = nullptr;
 
 QueueRegistry&
 QueueRegistry::get()
 {
   if (!me_) {
-    me_ = new QueueRegistry();
+    me_.reset( new QueueRegistry() );
```
eflumerf, line 25: One of the "modernize" guidelines recommends using std::make_unique in this situation.

test/FakeDataConsumerDAQModule.cpp
```diff
@@ -28,6 +28,9 @@ namespace dunedaq::appfwk {
 FakeDataConsumerDAQModule::FakeDataConsumerDAQModule(const std::string& name)
   : DAQModule(name)
   , thread_(std::bind(&FakeDataConsumerDAQModule::do_work, this, std::placeholders::_1))
+  , nIntsPerVector_(999)
+  , starting_int_(-999)
+  , ending_int_(-999)
   , queueTimeout_(100)
   , inputQueue_(nullptr)
```
eflumerf, line 35: Should we move initializers for basic-types to the header, using brace syntax?

test/FakeDataProducerDAQModule.cpp
```diff
@@ -29,6 +29,11 @@ FakeDataProducerDAQModule::FakeDataProducerDAQModule(const std::string& name)
   , thread_(std::bind(&FakeDataProducerDAQModule::do_work, this, std::placeholders::_1))
   , outputQueue_(nullptr)
   , queueTimeout_(100)
+  , nIntsPerVector_(999)
+  , starting_int_(-999)
+  , ending_int_(-999)
+  , wait_between_sends_ms_(999)
```
eflumerf, line 35: Should we move initializers for concrete types to the header, using brace syntax?

test/FakeDataProducerDAQModule.cpp
```diff
@@ -106,7 +111,11 @@ FakeDataProducerDAQModule::do_work(std::atomic<bool>& running_flag)
     oss.str("");
 
     TLOG(TLVL_TRACE) << get_name() << ": Pushing vector into outputQueue";
-    outputQueue_->push(std::move(output), queueTimeout_);
+    try {
+      outputQueue_->push(std::move(output), queueTimeout_);
+    } catch(const QueueTimeoutExpired& ex) {
+      TLOG(TLVL_TRACE) << "Push resulted in a timeout in FakeDataProducerDAQModule::do_work; data may be lost!";
```
eflumerf, line 117: Maybe ers::warning(ex) here?

test/queue_IO_check.cxx
```diff
@@ -242,6 +246,14 @@ main(int argc, char* argv[])
     return 1;
   }
 
+  if (vm.count("nelements")) {
+    nelements = vm["nelements"].as<int>();
+
+    if (nelements <= 0) {
+      throw std::domain_error("# of elements must be a positive integer");
```
eflumerf, line 253: throw std::... is discouraged; we might want to concoct and throw an ERS Issue here

test/queue_IO_check.cxx
eflumerf: Maybe a better comment to make would be to remove some of the defaults (queue_type?) and force the user to specify things via the command-line...But that could also be a separate Issue.

jcfreeman2: Happy to make it 1 million.

src/QueueRegistry.cpp
eflumerf: Except that it doesn't build. Rescinding comment.

test/queue_IO_check.cxx
jcfreeman2: Changed it to 1 million.

test/FakeDataConsumerDAQModule.cpp
jcfreeman2: After Slack discussion, agreed that parameters which are only ever meant to be settable via JSON should be set in the header.

test/FakeDataProducerDAQModule.cpp
jcfreeman2: After Slack discussion, agreed that parameters which are only ever meant to be settable via JSON should be set in the header.

test/FakeDataProducerDAQModule.cpp
jcfreeman2: Done.

test/queue_IO_check.cxx
jcfreeman2: Replaced the throws of std::domain_error with 
```
ERS_DECLARE_ISSUE(appfwk,                 ///< Namespace                                                                  
                  ParameterDomainIssue,  ///< Issue class name                                                            
                  "ParameterDomainIssue: \"" << ers_messg << "\"",
                  ((std::string)ers_messg))
```

Issue #99, opened 2020-08-17
A couple of fixes


jcfreeman2:
This Issue covers a couple of basic fixes which can be made to the appfwk code:

a) While appfwk v1.1.0 was cut back on August 5, it remains the case that on the develop branch, the CMakeLists.txt file has the version as v1.0.0, i.e. `project(appfwk VERSION 1.0.0)`. This should be fixed.

b) At some point back in May when the queue interface was being changed, the ability of queue_IO_check to allow you to test the behavior of a single thread popping elements off a queue without any threads simultaneously pushing elements onto it was lost. This should be reinstated. 

c) When I updated the CMake code to account for Pengfei's installation of nlohmann_json as a ups product, I removed the logic whereby the nlohmann_json header would be downloaded if the product wasn't found in the CMakeLists.txt file. However, I didn't do the equivalent for appfwkConfig.cmake.in, which I should.

jcfreeman2:
Marco reviewed the three pull requests associated with this issue, and they've been merged into develop.

PR #101, opened 2020-08-17
JCF: Issue #99: reinstate pop-only testing

jcfreeman2:
The main purpose of this commit is to allow queue_IO_check to test the
behavior of a single thread popping elements off a queue without any
threads pushing on to it at the same time. A couple of other
improvements are also made, however:

* Elimination of obsolete comments

* Clearer output for the tester (e.g., per-thread start time and execution time)

* Basic checks for the internal consistency of input (e.g., it's an
  error if you try performing a pop-only test and request that the queue start out empty)

### DIFF
--- a/test/queue_IO_check.cxx
+++ b/test/queue_IO_check.cxx
@@ -2,10 +2,10 @@
  *
  * @file queue_IO_check.cxx
  *
- * A low-level test of queue classes which inherit both from
- * QueueSource and QueueSink where we have a user-settable number
- * of threads writing elements to a queue while a user-settable
- * number of threads reads from the queue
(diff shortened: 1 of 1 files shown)


jcfreeman2:
A simple test would be to start the queue out with 10 million elements, and then pop them off:
```
queue_IO_check --pop_threads 1 --push_threads 0 --initial_capacity_used 0.01 --nelements 10000000 
```
...where the default capacity is 1 billion elements, so that 0.01 => 10 million initial elements. You can also see what happens if you change `--initial_capacity_used 0.01` to `--initial_capacity_used 0.00` (or just not set it at all)

mroda88 (approved):
the test is now more flexible.

Issue #106, opened 2020-08-30
const-qualifier of NamedObject member string breaks move operations


jcfreeman2:
Shortly before the v1.0.0 release, I modified the `std::string name_`  member of `NamedObject` by making it `const`. However, this breaks the move-ability of any class inheriting from NamedObject. While these lines
```
NamedObject(NamedObject&&) = default;
NamedObject& operator=(NamedObject&&) = default;
```
do compile (we'd have noticed if they didn't!), if a developer tries to move an object whose class inherits from NamedObject they'll be informed that the move constructor or assignment operator has been implicitly deleted. 

I'll fix this, and check to see if there are other places in appfwk where intended move semantics are broken. This is also a good opportunity for fleshing out the unit tests we use (a more thorough set, where we move a named object around, would have caught this at the compilation stage).

jcfreeman2:
Removed the const qualifier and added a unit test. To run the unit test, after you've checked out `jcfreeman2/issue106_namedobject_doesnt_move` and built, just do:
```
./build_daq_software --unittest`
```
and if you look at the tests for NamedObject you shouldn't see any errors. A move assignment error should appear, however, if you (locally!) revert the commit 5b34f59.

jcfreeman2:
With Eric's review, branch has been merged into develop.

PR #107, opened 2020-09-01
Jcfreeman2/issue106 namedobject doesnt move

jcfreeman2:
More-or-less a cut and paste from what I wrote for Issue #106:

Removed the const qualifier for NamedObject's name_ member so that it's move-assignable,  and added a unit test to test this (as well as its other move/copy semantics). To run the unit test, after you've checked out `jcfreeman2/issue106_namedobject_doesnt_move` and built, just do:
```
./build_daq_software --unittest`
```
and if you look at the tests for NamedObject you shouldn't see any errors. A move assignment error should appear, however, if you (locally!) revert the commit 5b34f59.


eflumerf (approved):
Tests pass in branch, fail when change is reverted, as described.

Issue #108, opened 2020-09-03
Make move semantics explicit for classes


jcfreeman2:
A couple of classes (DAQSink, DAQSource, DAQModule) don't follow the guideline that classes which contain member data have their move and copy semantics made explicit (Section 4.3, "Copyable and Movable Types", https://github.com/DUNE-DAQ/styleguide/blob/develop/dune-daq-cppguide.md). Of course, that doesn't mean they don't *have* copy and move semantics, just that they're set by the compiler based on the semantics of base classes, data members, etc. 

Marco's fine with making DAQModule explicitly non-movable and non-copyable (it's currently implicitly non-copyable and implicitly movable). Alessandro's fine with whatever choice of move semantics I come up with for DAQSink and DAQSource; I'll delete move semantics just because if we change our mind we won't break other people's code, while the converse isn't the case.

PR #109, opened 2020-09-03
JCF: Issue #108: make the copy/move semantics of DAQModule, DAQSink a…

jcfreeman2:
…nd DAQSource explicit


mroda88 (approved):
Perfect, this should fix the problem we had and make clear these object lifetime is non-standard

PR #112, opened 2020-10-05
Add schema-based / code generted command objects

brettviren:
This is a first PR for the schema-based, code-geneated command object support.  A few accompanying comments:

- There are some known lint errors in the codegen'ed code.  They are cosmetic and will be addressed either in a follow-up commit or in a future PR.  And at least one subsequent PR can be expected to provide some features wanted for the next release. (eg, see last bullet)

- There is a corresponding branch to `listrev` that depends on this branch and I'll make a PR there once things start to look okay here.  But, it may be useful to look at it w.r.t. this PR.  Once this PR is in, any modules will need a few simple changes like what is shown in this `listrev` branch.

  - https://github.com/DUNE-DAQ/listrev/tree/bv/config-proto

- There is some documentation for developers to make schema for their modules.  It still needs some work to both simplify and expand.

  - https://brettviren.github.io/moo/dunedaq-appfwk-schema.html

- Some design/reference documentation on how the codegen system works under the hood is here:

  - https://brettviren.github.io/moo/oschema.html

- As-is, this PR relies on the developer manually running the `schema/generate.sh` script to regenerate the code and that code is committed to the respoitory.  In a subsequent PR, we want to have the codegen run as part of the build and to not commit the generated files and need help from CMake experts.  Some initial investigation into how to do that with CMake is documented at the URL below.  But, that technique is not yet sufficient as CMake only sees the intermediate dependencies at CMake configuration time while it would be best that any of their changes are noticed at CMake build time.

  - https://brettviren.github.io/moo/buildsys.html#cmake


include/appfwk/CommandFacility.hpp
```diff
@@ -38,69 +38,42 @@
...
-   * @brief Singleton pattern; get a handle to the CommandFacility
-   * @return Reference to the CommandFacility
-   */
-  static CommandFacility& handle()
-  {
-    if (!handle_)
-      handle_.reset(new CommandFacility());
-    return *handle_;
-  }
+  CommandFacility(std::string /*uri*/) {}
```
eflumerf, line 49: It may be a TODO item, but I think it would be good to maintain Doxygen comments in the code

include/appfwk/DAQModule.hpp
```diff
@@ -67,17 +70,23 @@ namespace appfwk {
 class DAQModule : public NamedObject
 {
 public:
-  /**
+
+  using data_t = nlohmann::json;
```
eflumerf, line 74: I might suggest "json_t" or "config_t" for clarity, if we have to have a using...

eflumerf: I did also find a few places where nlohmann::json was used alongside data_t, and where data_t was used in files that didn't clearly inherit from DAQModule (for example, Logger.hpp). Maybe we could move the using to NamedObject.hpp if it's something we want to have broader visibility?

brettviren: Yes, I think some centrally defined `typedef` or `using` would be best!

As for a type name itself, it's hard for me to find a "perfect" one.  Some I've used or considered:

- `json` is what nlohmann suggests.  I am not fond because instances are not themselves really much to do with JSON per se (eg, they are not text, can be serialized to non-JSON text)
- `data_t` a bit generic even if inside the `appfwk::` namespace.
- `object_t` maybe my favorite, but instances can be arrays, numbers, strings, etc. so not totally apt
- `config_t` too specific.  instances are used also for "init", for command as a whole, for payload, for module-specific sub-payload.
- `command_t` too specific and too generic (same eg as above)
- `dynobj_t` (dynamic object) maybe apt, but a little cryptic

Maybe if no clear winner we stick with what nlohmann suggests?

BTW, `daq_application.cxx` is another spot outside of `NamedObject` that uses this type.

Once settled, I'm happy to push an update to this PR or to make fixes in a follow up.

glehmannmiotto:
How about cmddata_t?

Giovanna

———————-
CERN EP/DT-DI


On 6 Oct 2020, at 23:10, Brett Viren <notifications@github.com> wrote:

﻿

@brettviren commented on this pull request.

________________________________

In include/appfwk/DAQModule.hpp<https://github.com/DUNE-DAQ/appfwk/pull/112#discussion_r500597041>:

> @@ -67,17 +70,23 @@ namespace appfwk {
 class DAQModule : public NamedObject
 {
 public:
-  /**
+
+  using data_t = nlohmann::json;


Yes, I think some centrally defined typedef or using would be best!

As for a type name itself, it's hard for me to find a "perfect" one. Some I've used or considered:

  *   json is what nlohmann suggests. I am not fond because instances are not themselves really much to do with JSON per se (eg, they are not text, can be serialized to non-JSON text)
  *   data_t a bit generic even if inside the appfwk:: namespace.
  *   object_t maybe my favorite, but instances can be arrays, numbers, strings, etc. so not totally apt
  *   config_t too specific. instances are used also for "init", for command as a whole, for payload, for module-specific sub-payload.
  *   command_t too specific and too generic (same eg as above)
  *   dynobj_t (dynamic object) maybe apt, but a little cryptic

Maybe if no clear winner we stick with what nlohmann suggests?

BTW, daq_application.cxx is another spot outside of NamedObject that uses this type.

Once settled, I'm happy to push an update to this PR or to make fixes in a follow up.

src/FileCommandFacility.cpp
```diff
@@ -0,0 +1,262 @@
...
+#include <memory>
+
+using namespace dunedaq::appfwk;
+
+using object_t = nlohmann::json;
+
+// Abstract the object level
+struct ObjectStream {
+    std::string name;
+    std::iostream& io;
```
mroda88, line 28: This is a bit dangerous, as it does not control the lifetime of the stream given in constructions. 
Can we do something safer?

src/FileCommandFacility.cpp
```diff
@@ -0,0 +1,262 @@
+/** @file FileCommands.cpp
+
+    Fixme: the local "streams" types supporting FileCommandSource are
+    general and could be useful broken out.
+
```
mroda88, line 5: Why this is file does not have a dedicated header? 
It even contains a virtual class. Otherwise I don't see the reason of the hierarchy.

src/DAQModuleManager.cpp
```diff
@@ -0,0 +1,133 @@
...
+    auto cmd = cmd_data.get<cmd::Command>();
+    ERS_INFO("Command id:"<< cmd.id);
+
+    if (cmd_data.contains("waitms")) {
+        int waitms = cmd_data["waitms"];
+        std::this_thread::sleep_for (std::chrono::milliseconds(waitms));
+    }
+
+    if ( ! initialized_ ) {
+        if ( cmd.id != "init" ) {
```
mroda88, line 121: Hardcoding this does not seems a good idea to me. We are hardcoding commands.

src/FileCommandFacility.cpp
```diff
@@ -0,0 +1,262 @@
...
+            try {
+                command = this->recv();
+            }
+            catch (const StreamExhausted& e) {
+                ERS_INFO("Command stream end");
+                break;
+            }
+
+            manager.execute(command);
```
mroda88, line 250: This needs a try and catch loop here. ModuleManager can throw and we are not catching it.

mroda88 (commented):
There are a few open questions not clear from the code.
* The command facility is not a singletone anymore. Is there a specif reason for this? Do we envision multiple command facilities active on the same ModuleManager? or do we expect multiple command facilities acting on different module managers in the same application?
* I don't really understand why we renamed the DAQProcess to ModuleManager

CMakeLists.txt
```diff
@@ -18,12 +17,13 @@ set(APPFWK_DEPENDENCIES ${CETLIB} ${CETLIB_EXCEPT} ers::ers Folly::folly nlohman
...
 
-add_library(appfwk SHARED src/QueueRegistry.cpp src/DAQProcess.cpp src/DAQModule.cpp)
+add_library(appfwk SHARED src/QueueRegistry.cpp src/DAQModule.cpp
+  src/DAQModuleManager.cpp src/DAQModuleHelper.cpp)
 target_link_libraries(appfwk ${APPFWK_DEPENDENCIES})
 target_include_directories(appfwk PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include> $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}> )
 
-add_library(appfwk_QueryResponseCommandFacility_duneCommandFacility src/QueryResponseCommandFacility.cpp)
-target_link_libraries(appfwk_QueryResponseCommandFacility_duneCommandFacility appfwk)
+# add_library(appfwk_QueryResponseCommandFacility_duneCommandFacility src/QueryResponseCommandFacility.cpp)
```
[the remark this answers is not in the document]
alessandrothea: Fine.

CMakeLists.txt
```diff
@@ -65,8 +68,8 @@ target_link_libraries(appfwk_FakeDataProducerDAQModule_duneDAQModule appfwk)
 add_executable(queue_IO_check test/queue_IO_check.cxx)
 target_link_libraries(queue_IO_check ${Boost_PROGRAM_OPTIONS_LIBRARY} appfwk)
 
-add_executable(dummy_test_app test/dummy_test_app.cxx)
-target_link_libraries(dummy_test_app appfwk appfwk_DummyModule_duneDAQModule ${Boost_PROGRAM_OPTIONS_LIBRARY} appfwk)
+# add_executable(dummy_test_app test/dummy_test_app.cxx)
```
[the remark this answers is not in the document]
alessandrothea: Fine.

apps/daq_application.cxx
```diff
@@ -109,22 +35,22 @@ ERS_DECLARE_ISSUE(appfwk,          // namespace
...
+
+  CommandLineInterpreter args;
+  try {
+    args = CommandLineInterpreter::parse(argc, argv);
+  } catch (ers::Issue& e) {
+    // Die but do it gracefully gracefully.
+    // Use of std::cout annoys the linter. 
+    std::cout << "Command-line parsing failed. Error:" << std::endl;
+    std::cout << e.message() << std::endl;
+    exit(-1);
```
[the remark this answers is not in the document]
alessandrothea: Noted. Frankly this is doesn't seem to be particularly relevant.

CMakeLists.txt
```diff
@@ -86,6 +89,9 @@ daq_add_unit_test(NamedObject_test		  )
 
 daq_install(TARGETS daq_application 
                     appfwk
-		    appfwk_QueryResponseCommandFacility_duneCommandFacility 
-                    VectorIntFanOutDAQModule_duneDAQModule StringFanOutDAQModule_duneDAQModule NonCopyableTypeFanOutDAQModule_duneDAQModule
+		    # appfwk_QueryResponseCommandFacility_duneCommandFacility
```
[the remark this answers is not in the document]
alessandrothea: Uh?

src/DAQModuleManager.cpp
alessandrothea: Noted. Consider this a placeholder for a future better way to ensure that the initialization command is received first.

CMakeLists.txt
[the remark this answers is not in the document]
brettviren: Are you asking about the commented-out QRCF, @mroda88?

In my git history I see this coming in through the start of @alessandrothea's branch but maybe that got picked up from where he started which may have been my branch.  In any case, let me give my view:

The idea of a QRCF is definitely not gone.  But since the whole notion of command handling changed, a QRCF needs a revamp.  There are a few ways to provide it and it's on my list to try some out and see what might work well.  Some approaches:

0) the file command facility (FCF) can take a sequence of commands, each with an optional time delay to "play" the sequence at some pace.  This is not at all interactive like QRCF, but can be used now.

1) I think a user can (almost?) use the FCF now with individual command files via `cat cmd.json > fifo.jsonstream` to deliver one command, where the file here is made via "mkfifo".  Then, the user can have a family of `.json` files ready and use the shell command line as sort of a cheap REPL.  I say "almost" because I need to test this to see how the FCF handles `cat` closing the fifo on each command.  Probably some C++ needs improvement to support this.

2) Essentially the same but instead of a fifo + the FCF the command would be delivered through @roland-sipos embedded HTTP server CF.

3) Provide support code in Python for defining the family of possible commands, use eg `ipython` or `ptipython` to provide bespoked REPL.

4) Of course, a REPL could be implemented in C++ like the previous QRCF.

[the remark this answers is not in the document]
alessandrothea: Yes, `QueryResponseCommandFacility` was commented out simply because in its current form can only provide the command and not the command-data payload.

include/appfwk/DAQModule.hpp
alessandrothea: Right, I think there is a general agreement that a centrally defined typedef would be beneficial.
I had a look around the code to remind myself how `nlohmann::json` objects are used in this PR, and they are all tied to the concept of command. In this new scheme even initialization data is distributed via commands, therefore it seems reasonable to consider `nlohmann::json` objects as "command data" in the current context.

I would then go for `cmddata_t`, as @glehmannmiotto suggested

CMakeLists.txt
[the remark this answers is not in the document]
alessandrothea: FYI: The command facility is scheduled to be moved in a different package as a standalone generic tool.

alessandrothea:
Thanks for the all the comments, @eflumerf @mroda88 
This PR is somehow different from the previous: it introduces substantial changes to the code and its the first of a few which will affect the command facility and the CMake files.

I suggest to proceed as follows:
- discuss and resolve comments about the PR structure in this PR, with setting a deadline to this evening (Thurs 8 Oct, US TZ)
- Merge the PR tomorrow morning (CERN time)
- Resolve any high-level code issues by tomorrow evening and prepare a follow-up PR
- Capture the low-level comments in a Github issue to ensure thy are not lost and are resolved before the next release.

In this way we will be able to speedly move on to the next changes (external command facility and such) and schedule a proper code review in the context of the daq-suite v2 release.

alessandrothea:
So, to get started:

> * The command facility is not a singletone anymore. Is there a specif reason for this? Do we envision multiple command facilities active on the same ModuleManager? or do we expect multiple command facilities acting on different module managers in the same application?

Well, basically because there was no specific reason to enforce the unicity of the command facility via a singleton pattern.
The instantiation of the command facility is responsibility of the `daq-application`, which is a natural place where to enforce uniqueness

> * I don't really understand why we renamed the DAQProcess to ModuleManager

OK, so this requires some explanation indeed. In absence of any service, the former `DAQProcess` ended up being effectively, a `DAQModuleManager` and, in general, even when services will be introduced, a use for a Module+Queue container remains. The idea is to re-introduce the `DAQProcess` at a later stage, when the first services will become available. The `DAQModuleManager` will remain, as a `DAQProcess` data member.

mroda88:
Hi all, in this particular conditions, I'm happy to merge code with room for improvements so that further and related developments are not stalled. So this approach seems reasonably fine for me but I think we miss too much of the big picture here to be able to address high level issues properly. 
The reason why I feel there is something missing is that from here we only see the changes for the AppFwk code. I thought a bit about what I miss and I think it can all be summarised as "How the creation of new DAQModules changes as a function of these changes?" 
I can give some more specific examples:
* How many files do we have to add for each new DAQModule? 
* Where can we put code that ensures that a configuration is correct? Is this code supposed to be written by the user? If not, can we tell the template to add more checks apart from formal checks? This is a doubt I have following the fact that some checks are still hardcoded in some DAQModule code (for example FakeDataConsumerDAQModule::init), but from the presentations I thought that the validation of the schema was done before sending the configuration to a process, not by the Module itself. 
* How much redundancy is expected when we write code using this system? This is a question that rises looking at the code that handles the queue types that is essentially duplicated. Is it something we will be able to optimise? 

I had a look at the links attached to the pull request but they mostly focus on technical details.  I honestly would love (before discussing the high level details) at least an hint on [How to create your own DAQModuel](https://github.com/DUNE-DAQ/appfwk/wiki/Step-by-step-instructions-for-creating-your-own-DAQModule-package-under-v1.1.0) will change in the new system.

alessandrothea:
@brettviren you want to take this?

brettviren:
Hi @mroda88 . These are really good questions.  There's still work needed to streamline some things though much of that will be done outside of appfwk C++, per se.

I'll try to give some specific answers from my view of things:

> How many files do we have to add for each new DAQModule?

Ultimately, each `DAQModule` implementation needs to provide a little schema data structure.  It would go in a single file.  If a package provides many `DAQModule` implementations, their schema can be kept in separate files or rolled up into one.

> Where can we put code that ensures that a configuration is correct?

The schema files go under `<pkg>/schema/`.   They will be installed by CMake to some central location (tbd, but maybe `$CMAKE_INSTALL_PREFIX/share/schema` ?)

We currently commit the generated C++ headers and this is not "correct" to do.  After we get codegen integrated into CMake we will remove and cease committing them.

> Is this code supposed to be written by the user?

Just the module level schema structure.  Ie, the same content they might otherwise spread throughout their module class gets concentrated into the schema.

The rest will be either automatically derived from those little schemas or from some general, hand-written support code (not the module developer's problem).  For example:

- The codegen'ed C++ structs and `nlohmann::json` serialization functions already and automatically enforce a strong level of validation.  This is maybe what you mean by:

> If not, can we tell the template to add more checks apart from formal checks? 

In addition, we can explore extending the C++ codegen for example to bake in constraints that are given in the schema (numeric ranges, string patterns, preferring enum over free strings).

Continuing with your comment:

> This is a doubt I have following the fact that some checks are still hardcoded in some DAQModule code (for example FakeDataConsumerDAQModule::init), but from the presentations I thought that the validation of the schema was done before sending the configuration to a process, not by the Module itself.

Yes, correct.  @alessandrothea and I argued back and forth about this and it probably needs still more discussion than can easily by typed out here.  There are a few queue info access patterns that our current modules need.  I might call them: **named** (I want "this particular queue"), **all** ("I want all the queues"), **matching** ("I want all queues of some characteristic, eg 'input'").  Providing for these patterns (and maybe new ones!) in the single common schema which does not require the module to do manual validation is a goal we have not yet quite met.

- We have the start of Python code derived from schema that allows creation of configuration objects with "valid by construction" methods.  Ie, the Python itself provides some guarantee that the object it produces has some validity.  We also have some mock-up doing this directly in Jsonnet.

- Whole-command object validation via JSON Schema.  `moo` does this already by building JSON Schema from our schema files and applying validation to a constructed command object.  Still to do is to automate the construction of the whole-command object schema where as for now some manual work is needed at the job level (not module level).

- Eventually we want to roll a lot of this into some kind of "configuration editor".

> How much redundancy is expected when we write code using this system? This is a question that rises looking at the code that handles the queue types that is essentially duplicated. Is it something we will be able to optimise?

This is also a really good question and a keen observation.  We have talked about also moving this module-level queue "endpoint" initialization into codegen and I think it's still a good idea.  This code is essentially parameterized the "endpoint" types (eg `QueueSink`), the queue name and the access pattern (again, named/all/matching).  And, since almost all that info is in the schema now, this is certainly ripe for being provided by codegen.

But, how best to express the schema for this part, and what C++ patterns to generate (mixin class?  a struct held as module data member?) just hasn't been sussed out.

So, for right now, we do require some repetitive coding by module developers.  I've done this but only for `appfwk` and `listrev` modules.

CMakeLists.txt
```diff
@@ -1,8 +1,7 @@
 cmake_minimum_required(VERSION 3.12)
 project(appfwk VERSION 1.1.0)
 
-set(CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/../daq-buildtools/cmake ${CMAKE_MODULE_PATH})
-include(DAQ)
+find_package(daq-buildtools)
```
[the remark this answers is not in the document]
jcfreeman2: Good question. appfwk will build with the find_package call if you're using the head of daq-buildtools' develop branch (i.e., if you follow the instructions at https://github.com/DUNE-DAQ/appfwk/wiki/Compiling-and-running). However, I asked Alessandro a couple of days ago if the expectation should be that a package build against the head of daq-buildtools' develop branch or against the most recent public frozen release of daq-buildtools, and he said it should be against the most recent release (i.e., the instructions at https://github.com/DUNE-DAQ/appfwk/wiki/Compiling-and-running-under-v1.1.0). So, for the time being you'll want to leave the original lines in, but when the next daq-buildtools release comes out - O(1 week) - the new line will work.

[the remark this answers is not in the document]
jcfreeman2: @mroda88 @alessandrothea

mroda88:
Thanks @brettviren this reassures me quite a lot as it means I understood your presentations. The code as it is was giving me the idea of a less powerful system with respect to your presentations. I really look forward to a scenario where this will be implemented, it will be very elegant. The user prospective still deserves a bit of thoughts. Would you be able to add a draft of Step-by-step instructions for creating your own DAQModule package under v2.0.0 where at least we list what is going to change. I guess the how will depend on the evolution of AppFwk and the rest of the repos of the suite, but at least the things that will change would be nice to have them pointed out.

mroda88 (approved):
ok, so following the comments I'm approving the pull request assuming that the issues will be addressed once the surrounding pieces of code (CCM and CommandFacility) will become more stable. 
The documentation issue is still a **major** problem as it shows that we don't have a clear roadmap to offer to the users to help them build their Modules. I hope we will compile this together as a group while we decide what is the best approach for the users that also offers us an easy maintainability of the code.

alessandrothea (approved):

alessandrothea:
Thanks @mroda88. I've fixed a couple of details that were preventing the code to build in the daq build environment and hit the button merge.

PR #113, opened 2020-10-12
Feature cmdlib

roland-sipos:
This is the PR for removing the CommandFacility that was moved to become a more generic interface in cmdlib:
https://github.com/DUNE-DAQ/cmdlib

Main items in the PR:

- Removed `CommandFacility` and `fileCommandFacility` (also related declared issues and targets in CMake.)
- cmdlib is a library that the appfwk depends on
- `DAQModuleManager` inherits `cmdlib::CommandedObject` for implementing `execute`
- Updated the daq_application to resemble a use-case how the cmdlib library is used

For further details please have a look on cmdlib and its examples.

There is also a `restCommandFacility` implementation that uses HTTP REST for receiving commands. One can send command JSON objects via curl or with the python example application under:
https://github.com/DUNE-DAQ/restcmd

Any questions and suggestions are welcome.


eflumerf:
Purely from an appfwk standpoint, I don't see anything wrong with the changes in this PR. There may be some integration issues in the cmdlib repo, but I think that we should open Issues against that repository to address them, not hang up this PR.

eflumerf (approved):

brettviren:
I feel this change would create an inversion of proper dependency.

Frameworks define interface base classes whose implementations are supplied in other packages.  The framework package does not depend on those implementation packages. 

I think `appfwk` should continue to provide the base `CommandFacility` interface class.  

If `cmdlib` is meant to provide some non-`appfwk` code, that's certainly fine.  In that case, an `appfwk-cmdlib` package may be used to hold code which depends on both packages.

roland-sipos:
> I feel this change would create an inversion of proper dependency.
Usually software frameworks use a set of services, in which case there is no need for the services to be encapsulated in the framework code itself. I think any new "target" directly in the build of framework code naturally implies that the service became part of the appfwk. How the "command reception" and "callback on execution" is a form of service library, ers is another one for error reporting. Also, we shouldn't forget that there will be (even if a minimal) set of non-appfwk DAQ applications that still need to be commanded by the same infrastructure.

> Frameworks define interface base classes whose implementations are supplied in other packages. The framework package does not depend on those implementation packages.
This is definitely true for features that are provided by the framework. The "command service", how I understood, was a CCM enforced interface via a library that the appfwk uses. (In this case, implementing commanded objects. e.g.: `DAQModuleManager`)

PR #117, opened 2020-10-23
Moo updates

brettviren:
This adds copies of the "struct" and "nljs" templates from moo so they can be more easily customized to what details we want for `appfwk`.  It updates `generate.sh` to work with latest `moo` and includes a regen.

### DIFF
--- a/include/appfwk/cmd/Structs.hpp
+++ b/include/appfwk/cmd/Structs.hpp
@@ -66,7 +66,8 @@ namespace dunedaq::appfwk::cmd {
     using InstName = std::string;
 
     // @brief Capacity of a queue
-    using QueueCapacity = uint64_t;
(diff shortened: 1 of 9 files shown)


eflumerf (approved):
Of course, the are generated files as part of this commit, which it would be really nice to have generated through CMake. Is this an issue somewhere? I don't see it in the appfwk issues, but there might be one in cmdlib...

brettviren:
@eflumerf yes, this is a recognized problem and accepted (for now) as a temporary state of affairs.

This issue is related to moving the codegen into cmake.  https://github.com/DUNE-DAQ/daq-buildtools/issues/47

eflumerf:
Excellent, thanks for including the link to the issue!

PR #119, opened 2020-11-09
(more) sensible handling of cetlib plugin errors when creating modules

alessandrothea:
`DAQModule`'s make wrapper now catches the `cet::exception` thrown when the requested plugin is not found and wraps it into an ers issue.


mroda88 (approved):
ok

PR #121, opened 2020-11-16
Fix #120 by baking TLAs and grafts into a per-schema model file.

### LINKED ISSUE #120: Bake specifics into a schema-specific model file.
Let's remove the need to have CMake "know" about moo TLAs by baking them into a per-schema model file.  See previous discussion at:

https://github.com/brettviren/moo/issues/5

brettviren:
Note, the generated files end up exactly identical as before so this
does not change any code.  It just sets us up for simpler CMake
integration.

### DIFF
--- a/schema/appfwk-cmd-model.jsonnet
+++ b/schema/appfwk-cmd-model.jsonnet
@@ -1,0 +1,17 @@
+// This is a "model" file specific for the "cmd" schema of the appfwk.
+//
+// It's job is to provide the glue between the schema data structure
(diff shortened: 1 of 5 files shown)


jcfreeman2 (approved):
I've confirmed that after Brett's shifting the setting of the variables from commandline TLAs to settings inside the model files, the Structs.hpp and Nljs.hpp files for the FakeDataProducer and FakeDataConsumer DAQ modules come out identical to what they were before when `generate.sh` is run

PR #122, opened 2020-11-17
Updating CMakeLists and directory structure to dbt v1.3.0

alessandrothea:
CMakeList updated to useing `daq-cmake`
Private headers in `test/include` moved to `test/src'


mroda88 (approved):
Tested and compiles for me.

PR #123, opened 2020-11-18
Revert "Fix #120 by baking TLAs and grafts into a per-schema model file."

### LINKED ISSUE #120: Bake specifics into a schema-specific model file.

alessandrothea:
Baking TLAs into per schema model files has duplicated the number of schema-related files in the develop branch to help with the development of the cmake layer. This goes against the rationale of pushing complexity into the build system rather than on developers. With the understanding that the cmake-codegen layer is still work in progress, it is preferable to wait before applying such changes to develop.

Reverts DUNE-DAQ/appfwk#121


jcfreeman2:
After discussing this PR with Alessandro this morning and modifying generate.sh to reflect the current layout for private header locations, I tested this by confirming that generate.sh still produces the headers we want in the locations we want.

jcfreeman2 (approved):
See my last comment; all looks well.

PR #124, opened 2020-11-24
Jcfreeman2/build against daq cmake issue1

jcfreeman2:
To be pulled only after daq-cmake's pull request number 7 has been pulled


dingp (approved):

dingp:
tested together with https://github.com/DUNE-DAQ/listrev/pull/9 and DUNE-DAQ/daq-cmake#7

PR #129, opened 2020-12-16
Commands handling by DAQModuleManager

### LINKED ISSUE #128: Commands handling by DAQModuleManager
Observed the following issues with the handling of commands in DAQModuleManager when testing dunedaq-v2.0.0:
1) The module manager doesn't check if a module supports a specific command. It shall use the DAQModule::has_command(commandName) method to check. Now if a command is not supported by a module this breaks the complete command for the application.
2) If a module fails the execution of a command the module manager aborts the command dispatching instead of completing the dispatching and then collecting information about all modules that failed.
3) If for a command a module doesn't have parameters, the command should still be invoked on the modules.

I'm going to post candidate fixes for those issues in a branch.

alessandrothea:
Implements the changes discussed in #128.
Resolves #128.


alessandrothea:
**Note**: not a particular clean implementation, but it seems to properly handle the concerns raised in #128.

glehmannmiotto (approved):
ok for now; in future we may revise parameters matching.

alessandrothea:
Re parameter matching: yes, of course. This PR is about having a simple and reliable recipe now, but poses no constraints to further evolutions.

PR #131, opened 2020-12-18
Ignore moo-generated files

philiprodrigues:
This PR adds the autogenerated Structs.hpp and Nljs.hpp files in test/src to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,9 @@ _deps
 
 #Emacs
 *~
+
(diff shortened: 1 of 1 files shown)


jcfreeman2 (approved):
Quite straightforward to test; I confirmed that after merging in Phil's fix "git status" no longer refers to the auto-generated headers. I'll look into updating .gitignore for other packages in the same spirit.

PR #138, opened 2021-01-22
Cleaning up jsonnet configuration generator scripts

alessandrothea:
Also moving the `FakeDataProducer` and `FakeDataConsumer` plugins to the `daqdemos` repository.


brettviren:
Just to leave a comment for future searchers, this provides the second time that the ddcmd/ddrepl docs get broken!

https://brettviren.github.io/dune-daq-repl/

I'm not against the merge, just propagating a heads up.

Issue #130, opened 2020-12-16
Styleguide compliance


mroda88:
The appfwk has still some issue with the styleguide. This is a reminder for us all to fix it if we notice something wrong.

PR #139, opened 2021-01-28
Rename variables and methods to conform to updated style guide.

eflumerf:
Clang-format all files.
Resolve linting issues.


eflumerf:
Resolves #130

bieryAtFnal (approved):
Several people have looked at these changes and confirmed that they look reasonable.  In addition, we have verified that this code runs correctly within a simple test of a partial minidaqapp.

PR #142, opened 2021-02-08
Eflumerf/daq module interruptible wait

eflumerf:
Add an `interruptible_wait` method to DAQModule, which can be aborted early by updating the state variable and calling `DAQModule::interrupt`.


roland-sipos (commented):
I think the whole concept of an interruptible wait implementation is far more useful than just using it in the DAQModule. (Especially until we don't move to a compiler with C++20's std::jthread.) 

@eflumerf also had a great remark about the `Interruptible` aspect being more fitting to an extension on the queue interface's pop/push calls.

So all in all, this is really promising. And after a quick chat with @eflumerf , I think this should be extracted to an `Interruptible` class. Then we can discuss how `Interruptible` should be inherited, or as an interface provide the means to register threads' bind functions to wait on the internal cond var.

roland-sipos:
The new changes look much better and usable for other use-cases. Also, sorry for being annoying, but an utility like this begs for a single example test_app to showcase how it can be used in another context. 

After a proper test case (e.g.: aggressive interrupts test app), this can be merged. Thanks @eflumerf .

roland-sipos (commented):
@eflumerf
I like much more the current separation of the `Interruptible` utility class.

On the other hand, I'm not entirely convinced about the use case... We need @alessandrothea and maybe other appfwk devs to understand where do we need interrupts.

alessandrothea:
Hi @eflumerf , I agree with @roland-sipos that a separated interruptible utility class makes more sense.
That said, I am still quite puzzled by how you propose to apply it to DAQModules and Queues.
Both DAQModules and Queues are objects with potentially multiple functions. So a single interruptible wait doesn't seem to make much sense e.g. a Queue has waits for pushes and pops, therefore it must be one or the other depending on the use case.

alessandrothea:
Looks good. Will merge soon

Issue #133, opened 2020-12-31
Exceptions crashing daq_application


jcfreeman2:
More precisely: exceptions which are thrown from functions wrapped by `appfwk::ThreadHelper` will end `daq_application`. If I understand correctly, we don't want user code to end `daq_application`s but want them to flag an error which Run Control can then handle. 

Seeing this in action is simple. E.g., you can build `listrev` as described in https://github.com/DUNE-DAQ/appfwk/wiki/Compiling-and-running-under-v2.0.0 but before running `list-reversal-app.json` set the number of ints produced by a vector to a negative number, e.g., 
```
sed -r -i 's/"nIntsPerList": 5/"nIntsPerList": -1/' sourcecode/listrev/test/list-reversal-app.json 
daq_application -c stdin://sourcecode/listrev/test/list-reversal-app.json
```
or for that matter you can do something similar with `appfwk/schema/fdpc-job.json`. In both cases an `std::bad_alloc` will sail out of `daq_application`, ending it. 

We should think of ways to make it as difficult as possible for DAQ module developers to crash `daq_application` with their code; it would be worth investigating what the best way to keep an `appfwk::ThreadHelper`'s function from doing so would be.

philiprodrigues:
Some possibilities:

* As in #145, we could wrap the user function with code that stores any exceptions in a `std::exception_ptr`. The `exception_ptr` can be used to re-throw the exception in `stop_working_thread()`
* A more flexible approach, suggested by Marco, would be to use `std::async` and get a `std::future<void>` back. We could pass the future to the caller of `start_working_thread()`, or have it available via some `get_future()` function. Caller then has the option to wait on the future, or call `future.get()` to have the exception re-thrown wherever caller desires. (I don't see a way to explicitly check whether a future is ready, but `wait_for` with timeout=0 would presumably achieve that). If the future has not been passed out of `ThreadHelper` via `get_future()`, `stop_working_thread()` could (optionally) rethrow the exception
* We could use either of the above techniques to prevent the exception crashing the entire application, but just never rethrow it?

mroda88:
Regardless of what is the way we use to retrieve the exception, in order to decide if rethrow or not, I guess we want to decide what we want to achieve with the rethrowing. Considering that the rethrowing will be triggered by the stop call, the exception will probably be very late to be useful. 
[...]

PR #145, opened 2021-02-12
Re-throw exceptions from the working thread in the thread that calls stop

philiprodrigues:
As discussed in #133, exceptions thrown from the working thread in `ThreadHelper` crash the application. I found a possible approach here using `std::exception_ptr`:

https://vorbrodt.blog/2019/03/24/propagate-exceptions-across-threads/

I've implemented this approach here. Any exceptions thrown in the working thread are rethrown in the thread that calls `stop_working_thread()`. I'm not very familiar with this technique, so more pairs of eyes would be useful. If we decide to go with the general approach, I will add more documentation.

### DIFF
--- a/include/appfwk/ThreadHelper.hpp
+++ b/include/appfwk/ThreadHelper.hpp
@@ -18,6 +18,7 @@
 
 #include "ers/ers.h"
 
+#include <exception>
 #include <functional>
 #include <future>
 #include <list>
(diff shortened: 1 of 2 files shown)


mroda88:
Hi Phil, I think this is exactly the same functionality that you get when you use `async` and `future`: once the future state is ready it also throws any exception. If we want to go in that direction I recommend to simply use those features and not re-invent the wheel. 
Still, for a time dependent system, would we have benefit in knowing at stop that something went wrong? would it not be too late?

philiprodrigues:
Yeah I read that you can do the same with async and futures, but I didn't look into it in detail, so I'm not sure about the relative pros and cons. Where/when would the exception get rethrown in that case?

mroda88:
The exception would be rethrown when future::get is called.

philiprodrigues:
Having read up a bit on futures, I agree that they would provide a more flexible interface. Let's move the discussion to the original issue #133

PR #146, opened 2021-02-15
Add ctors for DAQ(Source|Sink) that take the 'name', not the 'inst' of the queue

philiprodrigues:
The existing constructors for `DAQSink` and `DAQSource` take the queue instance name ("inst" in the config), rather than the name of the queue as seen by the module ("name" in the config). So module writers have to explicitly query the config to match their name with the instance name. This PR adds constructors to `DAQSink`, and to `DAQSource`, which do that querying for the user. They look like:

```cpp
DAQSource(const nlohmann::json& init_json, const std::string& name);
DAQSource(const cmd::ModInit& init, const std::string& name);
```

(The `ModInit` version just converts its argument to `nlohmann::json` and then does the same as the other ctor)

Before-and-after usage example, from `TriggerDecisionEmulator`:

https://github.com/DUNE-DAQ/trigemu/commit/f7d55372f31a4b8d76db03e5ae2fa8229c3caec8

I've left the old constructor as it is: it's needed in a few cases, such as `dfmodules` plugins which don't know at compile-time how many queue instances they'll be talking to. This means everything is backwards-compatible, but it makes the overall API a little confusing: `DAQSink(const std::string& inst)` takes the "inst", while the other two (new) constructors take the "name".

If there's agreement that something like this is useful, I'll add doxygen comments

### DIFF
--- a/include/appfwk/DAQSink.hpp
+++ b/include/appfwk/DAQSink.hpp
@@ -9,8 +9,12 @@
 #ifndef APPFWK_INCLUDE_APPFWK_DAQSINK_HPP_
 #define APPFWK_INCLUDE_APPFWK_DAQSINK_HPP_
 
+#include "appfwk/Issues.hpp"
 #include "appfwk/Queue.hpp"
 #include "appfwk/QueueRegistry.hpp"
+#include "appfwk/DAQModuleHelper.hpp"
+#include "appfwk/cmd/Structs.hpp"
+#include "appfwk/cmd/Nljs.hpp"
 
 #include "TRACE/trace.h"
 #include "ers/Issue.h"
@@ -40,6 +44,9 @@ public:
   using duration_t = std::chrono::milliseconds;
 
   explicit DAQSink(const std::string& name);
+  DAQSink(const nlohmann::json& init_json, const std::string& name);
+  DAQSink(const cmd::ModInit& init, const std::string& name);
+
   void push(T&& element, const duration_t& timeout = duration_t::zero());
   void push(const T& element, const duration_t& timeout = duration_t::zero());
   bool can_push() const noexcept;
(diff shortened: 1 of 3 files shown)


brettviren:
A minor comment: I've seen this exact pattern called an anti-pattern.  That is the pattern of passing to a function two unwanted objects so a third actually wanted object is produced inside the function followed by discarding the two original inputs.  The problem is that it binds the interface to unrelated types, forces caller to use those types (though here, you keep backwards compatibility) and tends to make the interface grow as more unrelated types are used to provide the type that is actually wanted.  Here we see already growing two methods to ultimately provide the same object.

The cake-and-eat-it-too solution would be to provide one or more intermediate builder functions to be called outside the context where its result is needed:

```C++
#include "appfwk/some_utility.hpp" // provides std::string queue_inst()
DAQSource ds(appfwk::queue_inst(init), name);
```

BTW, I agree both `inst` and `name` are too generic.  Maybe `queue_name` for `inst`?  Maybe `endpoint_name` or `queue_alias` for `name`?

philiprodrigues:
I'm not quite following your example. What would `queue_inst(init)` return here?

I like `inst` -> `queue_name` and `name` -> `endpoint_name`

brettviren:
Urg, sorry.  I got the parens wrong.  I mean:

```c++
DAQSource ds(appfwk::queue_inst(init, name));
```

The function would dig out the `inst`from the `init` based on `name` and return it.

philiprodrigues:
Got it. How about putting it in `DAQModuleHelper.hpp` where the implementation could just be:

```cpp
return queue_index(init, {"name"})["name"].inst;
```

philiprodrigues:
See #149

PR #159, opened 2021-03-25
Primitive RESTful API

Juravenator:
This change implements a basic RESTful API in appfwk.

A new class `dunedaq::rest::Api` has been created. 
It does not (yet) use the built-in OpenAPI documentation generator that is demonstrated [here](https://gitlab.eecs.umich.edu/basaksum/vc-qc/blob/0a2b8d55c432e5b584688250ae66e4f86d841915/pistache/examples/rest_description.cc). Rather it exposes more naive functionality that allows developers to associate a method+url (e.g. `GET /api/v0/abc`) to a handler function.

`dunedaq::appfwk::Application` creates an instance of this API and handles initialization. It is also demonstrates how other code areas in a DAQ application can enrich the API with more data and endpoints.

As a first version, it merely implements a basic z page `/api/v0/healthz`, another remnant of Borg.
Their usefulness is demonstrated [here](https://vimeo.com/173610242).

With this endpoint, we now have the required functionality for an RC system to no longer fly blind.

It introduces a new flag `--rest-api` denoting the port number the API should listen to. If this is omitted, it uses port 0 and lets the OS choose an available port. The actually used port is available via `dunedaq::rest::Api::get_port()`

The API also serves `/api/v0/modules` and `/api/v0/commands/history`. These endpoints describe the active modules and their available commands, and the history of executed commands and their payload.

As a convenience, the `/` endpoint prints a list of available endpoints
```bash
$ curl localhost:8080/              
{
  "routes": [
    "GET /",
    "GET /api/v0/commands/history",
    "GET /api/v0/healthz",
    "GET /api/v0/modules"
  ]
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.12)
 project(appfwk VERSION 2.2.2)
 
-find_package(daq-cmake REQUIRED )
+find_package(daq-cmake REQUIRED)
 
 daq_setup_environment()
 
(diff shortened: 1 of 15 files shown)


glehmannmiotto:
Sorry, Glenn, what is the relation of this with cmdlib and restcmd?

Juravenator:
The commandfacility implementations, including the RESTCommandFacility, are only able to handle commands. It is not made to do other fundamental things like health/status checks, or serve a full RESTful API.
This caused problems such as the RC currently 'flying blind' as mentioned in previous CCM meetings.

This API implements the basics consisting of a couple of zpages and allows run control to no longer fly blind. It is however made to be extendable, which could make it potentially overlap with functionality of RESTCommandFacility if desired.

Run Control can now probe `/healthz` to check if the application is ready to receive commands, and if it is in the state Run Control expects it to be.

alessandrothea:
Really to have missed the notification. It got lost in the daily flood of GitHub emails and I noticed only today after 15 days!

Thanks, this is a very interesting prof of concept. I like the possibility to inspect the list of modules, command history and the `/` endpoint serving the list of endpoints.

Now, I have just had a quick look at the code and didn't have the time to look at the videos yet, but as far as I understand the model proposed here is different from what has been used so far, where the command facility is the single entry point for the interactions between the application and its controller.
If we want to move away from that, which is certainly possible, we need to understand the requirements that drive the change on the RC front, as well as the how this new approach can fit in the application design. 
For example, what are the requirements to avoid flying bind? now and in the future with several hundred apps to keep an eye on? Can one keep a single interaction point between controller and application? How do the rest API, which provides access to some monitoring information, relates to operational monitoring?

This discussion seems best done in a CCM meeting, so may I suggest to ask Alex for a slot in one of the next?
Thanks!

eflumerf:
Doing some Github maintenance and happened upon this PR. I'm thinking that it should probably be closed due to age?

plasorak:
Actually, there are regularly some discussions about this, let's keep this branch and discussion if possible?

eflumerf:
@plasorak Once again, asking if this should be closed, or if we should have some more active discussions around it.

plasorak:
You can close it 😀

alessandrothea:
Sorry @plasorak , but why? Yes, we never found the bandwidth to talk about direct access api for DAQ applications. Until we find the time to do that this remains a valuable example, I think.

PR #166, opened 2021-04-14
Move BOOST_REQUIRE instaces in for loops out of main control path

eflumerf:


philiprodrigues (approved):
This looks good to me, and has the desired effect when I run it in my test release

Issue #163, opened 2021-04-07
User-oriented documentation


jcfreeman2:
If you can see this Issue, it means that the existing documentation for this package has already been added to the official documentation, https://dune-daq-sw.readthedocs.io/en/latest, via its placement in your repo's docs/ directory. The next step is to make sure that the documentation in docs/ tells your package's users (as opposed to developers) everything they need to know about the package. The goal is to get this done for the dunedaq-v2.4.1 release in a couple weeks.

For details/hints please see https://dune-daq-sw.readthedocs.io/en/latest/editing_package_documentation/.

jcfreeman2:
Marco and I just got rid of the existing `docs/*.md` files which covered material that's more properly dealt with through the daq-buildtools and daq-cmake documentation. Still to-do is for appfwk developers to write/update documentation covering DAQModules, queues, the concept of a `NamedObject`, the behavior of `ThreadHelper` (and how it's different than `std::thread`), etc.

bieryAtFnal:
Before anyone spends any time on the type of documentation that is described in the previous post, I think that it would be useful to have a short discussion about the different meanings of the word "user".

Sometimes, when we use the word "user", we mean someone who operates or tests DAQ systems.  Other times, we may refer to developers who write software that makes use of classes and methods from another software package as users of that other software package.

In my discussion with Alessandro and Giovanna a couple of weeks ago, my understanding was that the "user" documentation that we are talking about in this round of documentation is focused on "operator-users".  As such, there may be very little that is needed in the appfwk package.

alessandrothea:
Right, we did make a distinction between operational users and developers and agreed that the this first iteration should focus on the former for most packages (e.g. `minidaqapp`, or `readout` or `dfmodules`). 
Core packages and `appfwk` in particular are a different matter: all modules and apps developers rely on them therefore developers documentation must be provided.
So @jcfreeman2's request to document makes sense to me.

bieryAtFnal:
Sure, but where should developer documentation be provided?  At one point, I heard that this would be in Wiki pages, not the [official documentation system](https://dune-daq-sw.readthedocs.io/en/latest/).

jcfreeman2:
@bieryAtFnal : @alessandrothea can correct me if I'm wrong but I believe the answer depends on what's meant by "developer documentation". Specifically, which of the two categories of developer are we talking about:
* Developers of the package in question (e.g., the authors of `src/DAQModule.cpp`)? 
[...]

PR #167, opened 2021-04-20
Johnfreeman/issue163 remove non appfwk documentation

jcfreeman2:
I've stripped out the documentation from appfwk which is now covered by daq-buildtools and daq-cmake. Note that this only partly satisfies Issue #163 , since it doesn't add to or update the appfwk-specific documentation (how to write DAQModules, etc.)


mroda88 (approved):
The final result is working as expected.

PR #170, opened 2021-04-27
Queue monitoring

alessandrothea:
This PR introduces the monitoring of queue size and capacity. 

- Introduces a new `QueueBase` base class
- Implements a common `gather_info` method for all queues
- Add a `capacity` data member to Folly Queues and `DynamicUnboundedQueues` have a fuzzy concept of capacity


mroda88 (approved):
looks ok to me.

glehmannmiotto (approved):
The implementation looks good to me, in general.
The only change I would suggest is related to the can_push/can_pop that are reimplemented in StdDeQueue with the same implementation as the base class and with an equivalent implementation in FollyQueue.
The subdivision of which methods go into Queue and which in QueueBase is somewhat arbitrary, e.g. the can_push/can_pop could be also in QueueBase. On the other hand, it makes sense to keep those methods close to the push/pull.

alessandrothea:
Ah, yes good point about `can_push` and `can_pop`. Having had a look into the `DynamicBoundedQueue` code, it became clear that the logic in `can_push` and `can_pop` can be (much?) more complex than a comparison between a size estimate and the capacity. For instance `DynamicBoundedQueue` does not have a fixed capacity (i.e. one cannot get the number back from a  `DynamicBoundedQueue` object, or at least I couldn't) and the capacity value passed to the constructor is taken as indicative. The queue can go beyond the nominal capacity if deemed necessary by the queue itself.
This suggested to me that `can_push` and `can_pull` are best handled as specialized method and kept close to `push` and `pull`.

PR #173, opened 2021-05-01
JCF: Issue #163: remove all Markdown files copied from the Wiki which…

jcfreeman2:
… either (A) are directed at appfwk developers rather than users, (B) became obsolete at some point, (C) became redundant at some point, or (D) are more properly located in the Wikis/official documentation of other packages. 

Removal of a given Markdown file from the `docs/` directory should not be understood to imply that the documentation in question isn't useful, it may simply mean that the file has a more appropriate location (e.g., the appfwk Wiki for documentation directed at the developers of appfwk itself)


eflumerf (approved):
Knowing that the wiki is backed by a git repository, I'd also be in favor of deleting pages from the wiki which are obsolete...

jcfreeman2:
@eflumerf Agree that that's a good idea, though it'll of course require another set of decisions about which pages to eliminate than were used for this PR

PR #176, opened 2021-05-11
Johnfreeman/issue163 update documentation

jcfreeman2:


docs/README.md
```diff
@@ -1,20 +1,201 @@
...
+  
+  private:
+  
+     void do_conf(const data_t& conf_data);
+     void do_start(const data_t& start_data);
+     void do_stop(const data_t& stop_data);
+     void do_scrap(const data_t& scrap_data);
+};
+```
+A set of programming idioms developed over the first year of DAQ module development which, while not strictly necessary for implementing DAQ modules, have proven to be quite useful. They'll be described within the context of the functions above. Of course, you can also see them in action by looking at the source code of actual DAQ modules.
```
eflumerf, line 52: I think this sentence is talking about ThreadHelper, but it doesn't seem to belong here

docs/README.md
```diff
@@ -1,20 +1,201 @@
...
+
+Drilling down a bit deeper, the way that `daq_application`s can be configured so flexibly is that they are designed to have DAQ modules embedded in them, where DAQ modules are units of code designed to perform specific tasks and which pass data between each other by means of containers known as "queues". These specific tasks can vary widely; they include [producing fake data for testing purposes](https://github.com/DUNE-DAQ/readout/blob/develop/plugins/FakeCardReader.hpp), [putting data into long term storage](https://github.com/DUNE-DAQ/dfmodules/blob/develop/plugins/HDF5DataStore.hpp), and so forth. DAQ modules will typically execute user-defined functions when receiving standard transitions from Run Control: conf, start, etc. appfwk provides the `DAQModule` base class which users should derive their DAQ module class from in their own packages.
```
mroda88, line 13: The HDF5 example is wrong. HDF5DataStore is not a `DAQModule`, here it creates confusion. Use `DataWriter`.

docs/README.md
mroda88, line 52: After the first sentence of this paragraph it's probably useful to ass a link to our style guides.

docs/README.md
```diff
@@ -1,20 +1,201 @@
...
+
+For a JSON file which (among other things) defines queues, see [this example](https://github.com/DUNE-DAQ/flxlibs/blob/15e256c0df102b1fc93802e9ed79a7cfd8c0ea4a/test/felix_wib2_readout.json), where the two main things to define for a queue are (1) its capacity (the maximum number of elements it can hold) and (2) its implementation. The two primary queue options for DAQ running are "FollySPSCQueue" (Single Producer Single Consumer) and "FollyMPMCQueue" (Multiple Producer Multiple Consumer), both implemented originally for Facebook but found useful for DUNE.
```
mroda88, line 75: This paragraph might be misleading. The implementation is usually defined at the level of configuration, the implementor can only get it and specify the memory type of the queue via the template. I think it might be worth rephrasing this bit in a way that does suggest that it is the DAQModule code that specifies the queue implementation type.

docs/README.md
```diff
@@ -1,20 +1,201 @@
...
+```
+void MyDaqModule::do_conf(const data_t& conf_data)
+{
+  auto data = conf_data.get<mydaqmodule::Conf>();
+
+  m_calibration_scale_factor = data.calibration_scale_factor;
+  // ...and then set the other members which take per-configuration values...
+}
+```
+This of course raises the question: what _is_ `mydaqmodule::Conf`? It's a `struct`, but rather than being manually written the code for it is generated by the DUNE DAQ build system itself, using a `jsonnet` file as input. It's in the `jsonnet` file that the logical contents of the struct are defined; an example of this type of file can be found [here](https://github.com/DUNE-DAQ/listrev/blob/develop/schema/listrev/randomdatalistgenerator.jsonnet). Further details are beyond the scope of appfwk documentation and are instead covered in the [daq-cmake documentation](https://dune-daq-sw.readthedocs.io/en/latest/packages/daq-cmake/).
```
mroda88, line 89: The link to the daq-cmake documentation is great. But that link is purely technical. Here we need a sentence, or a paragraph, on the logic of the struct. 
We need to explain
1) this allows automatic checks on the variable retrieved by the module maximising compile time checks and reducing checks to be performed by the do_conf operation. 
2) this is the way we broadcast all commands, the do_conf function is simply the typical place where this mechanism is more evident to the implementor. But behind the scene they already touched this in the  `init` implementation.

docs/README.md
```diff
@@ -1,20 +1,201 @@
...
+
+### The `do_start` function
+
+Most DAQ modules are designed to loop over some sort of repeated action when the DAQ enters the running state, and it's in the `do_start` function that this repeated action begins. A very common idiom for the `do_start` function is, "Set an atomic boolean stating that we're now in the running state, and then start one or more threads which perform actions in loops which they break out of if they see that the atomic boolean indicates we're no longer in the running state".
```
mroda88, line 93: Following the comments from slack, I believe idiom is not a "difficult" word, but here I would use "procedure" or "approach". But probably this is an American vs British difference rather than an issue with non-native speakers.

docs/README.md
```diff
@@ -1,20 +1,201 @@
...
+### The `do_start` function
+
+Most DAQ modules are designed to loop over some sort of repeated action when the DAQ enters the running state, and it's in the `do_start` function that this repeated action begins. A very common idiom for the `do_start` function is, "Set an atomic boolean stating that we're now in the running state, and then start one or more threads which perform actions in loops which they break out of if they see that the atomic boolean indicates we're no longer in the running state". 
+
+While it's of course possible to accomplish this using the existing concurrency facilities provided by the C++ Standard Library, the appfwk package itself provides a class, `ThreadHelper`, which makes this easier. `ThreadHelper` is covered in detail [here](https://dune-daq-sw.readthedocs.io/en/latest/packages/appfwk/ThreadHelper-Usage-Notes/); when in use the `do_start` function can be as simple as follows:
+```
+void MyDaqModule::do_start(const data_t& /*args*/) {
+    m_thread.start_working_thread();  // m_thread is an `appfwk::ThreadHelper` member of MyDaqModule
+}
+```
```
mroda88, line 100: Here I would add that the ThreadHeper does not only help the implementor, but also the user (in the sense of shifter). Since the threadhelper makes renaming the thread easy, it makes the control of the operation easier. 
I would also add that because of that using ThreadHelper is preferred over STL solutions.
STL solution should be considered only if more advanced features other than those offered by ThreadHelper are required, e.g. future, promise, deferred launch, etc.

docs/README.md
```diff
@@ -1,20 +1,201 @@
...
+
+### The `do_stop` function
+
+Quite simple, basically the reverse of `do_start`:
+```
+void MyDaqModule::do_stop(const data_t& /*args*/) {
+    m_thread.stop_working_thread();  // m_thread is an `appfwk::ThreadHelper` member of MyDaqModule
+}
+```
```
mroda88, line 109: Here we need a comment about loops of data in the Application. 
In general because a module exchanges data we should drain the input queues, unless a module is connected to a graph whose stop behaviour is such that other modules drain the queue.
Also we should also say that the stop can allow a new start to be called. In either the start or the stop what needs to happen is that memory from previous run is removed. 
This last comment can go in the start description if we don't like it here.

mroda88 (approved):
Obviously I am annoying and I made a lot of comments. But this is great work John, thanks.

PR #190, opened 2021-07-02
Change the image name of the daq_application process

### LINKED ISSUE #185: Can we rename "daq_application" instances?
It might be nice if, instead of a number of "daq_application" instances in `top`, we changed the image name to the name given by the required `--name` parameter.

eflumerf:
Change the image name of the daq_application process to the application name given in the argument list (i.e. --name). Resolves #185


eflumerf:
We could also control this behavior via a flag option added to CommandLineInterpreter.

PR #192, opened 2021-07-14
Update daq_application.cxx

glehmannmiotto:
set/update the DUNEDAQ_PARTITION variable.

### DIFF
--- a/apps/daq_application.cxx
+++ b/apps/daq_application.cxx
@@ -76,7 +76,7 @@ main(int argc, char* argv[])
     exit(0);
   }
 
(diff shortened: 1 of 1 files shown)


eflumerf (approved):

apps/daq_application.cxx
```diff
@@ -78,6 +78,7 @@ main(int argc, char* argv[])
 
   // Set/Update the application name in the environment. Used by logging/ers.
```
eflumerf, line 79: Might be a good idea to update this comment

glehmannmiotto: done. thanks!

PR #196, opened 2021-08-20
Add script to run several configurations of queue_io_check.

eflumerf:
Fix path to JSON file now that dbt-build.sh auto-installs.

### DIFF
--- a/test/scripts/queue_io_check.sh
+++ b/test/scripts/queue_io_check.sh
@@ -1,0 +1,44 @@
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+
+function run_test() {
+
+# Run without options
+queue_IO_check $@
+ret=$?
+if [ $ret -ne 0 ]; then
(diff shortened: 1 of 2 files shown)


jcfreeman2:
Great to see this. Just a couple of quick comments:
* The failure check via `$?` should go right after the `queue_IO_check`, otherwise we're not testing the program's return value
* A shameless plug for those reading: this test app is described at https://github.com/DUNE-DAQ/appfwk/wiki/Queue-testing
* Some of the tests might benefit from having more than 1000 elements given how quickly they're moved in the absence of engineered pauses between pushes and/or pops
* If the intent of this script is not only to check that `queue_IO_check` doesn't undergo unexpected failures but also to perform actual benchmarking one thing to consider would be testing the time to push onto a queue with nothing popping, and vice-versa (this idea is mentioned in more detail on the wiki page)

eflumerf:
Queues have moved to `iomanager` with `feature/iomanager` (still in development). I have replicated this change on a branch in `iomanager`, and we can think about merging it there later.

PR #202, opened 2021-12-16
New configuration generation

plasorak:
Pretty much the same as: https://github.com/DUNE-DAQ/minidaqapp/pull/114


carloschavez:
This looks good to me after including the python package networkx.
The package is already requested to be included in the release.

PR #203, opened 2022-01-06
Lengthen timeouts used for Queue tests

eflumerf:


jcfreeman2:
Confirmed that with a 5 ms timeout rather than a 1 ms timeout (FollyQueue) or 2 ms timeout (StdDeQueue), the time it takes to attempt to pop off an empty queue no longer significantly overshoots the timeout (as was the case for StdDeQueue and FollyQueue).

jcfreeman2 (approved):

PR #204, opened 2022-01-13
Re-enable build of ThreadHelper_test

eflumerf:


eflumerf:
This PR is obsolete. ThreadHelper_test is now WorkerThread_test in utilities. (And is enabled in the build there)

PR #211, opened 2022-02-08
Eflumerf/remove queue config

eflumerf:
Remove the QueueConfig helper struct, instead using confgen structs directly. Resolves a fixme comment.

Corresponding compatibility branches have been made in timinglibs, dfmodules, and trigger, but there may be other compatibility changes needed elsewhere.


bieryAtFnal:
I talked with Eric and learned that this PR will become obsolete with the upcoming Unified API changes, so I am closing it.

PR #215, opened 2022-02-28
Add new TRB parameter

jmcarcell:
for the connections between DQM and DF


jmcarcell:
Not needed anymore since `conf_utils.py` was moved

PR #218, opened 2022-03-08
use getenv_ifset for DUNEDAQ_ERS_DEBUG_LEVEL -- requires code in nano…

ron003:
…rc/cfgmgr.py OR DUNEDAQ_ERS_DEBUG_LEVEL in env

### DIFF
--- a/python/appfwk/conf_utils.py
+++ b/python/appfwk/conf_utils.py
@@ -479,7 +479,7 @@ def generate_boot(apps: list, partition_name="${USER}_test", ers_settings=None, 
             "DUNEDAQ_ERS_WARNING": ers_settings["WARNING"],
             "DUNEDAQ_ERS_ERROR": ers_settings["ERROR"],
             "DUNEDAQ_ERS_FATAL": ers_settings["FATAL"],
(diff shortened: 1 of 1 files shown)


ron003:
To test with and without the change to nanorc:

without the corresponding change to nanorc:
1. make a configuration export TRACE_FILE=<value_specified_in_daqconf_wiki_instructions>
2. do `tlvlSg 0x555ff; tlvls` to see 555ff in the 2nd column of numbers
3. run nanorc and there should be an error saying DUNEDAQ_ERS_DEBUG_LEVEL is not in the enviroment and no default is specified.
4. export DUNEDAQ_ERS_DEBUG_LEVEL=-1 and rerun nanorc and there should be no error.
5. do tlvls and see that the 555ff is replaced by 000ff

with the corresponding nanorc change:

6. unset DUNEDAQ_ERS_DEBUG_LEVEL and copy in the change to nanorc (cfgmgr.py), redo  `tlvlSg 0x555ff; tlvls` to see again 555ff in the 2nd column. Then rerun nanorc and this time there should be no error. Redo the tlvls and that the 555ff remains.

plasorak:
Ok yes, this seems to behave as expected.

plasorak (approved):

PR #219, opened 2022-04-06
Add the extra command

plasorak:
I don't know if this feature disappeared, or this is the right way of doing this, but it looks like the extra_commands argument in DAQModule wasn't generating the proper JSON. This PR fixes that.


philiprodrigues:
The code looks sensible to me. Can you suggest a way to test it?

plasorak:
Actually, I thought this was ok, but then I stumbled upon this: https://github.com/DUNE-DAQ/daqconf/blob/develop/python/daqconf/apps/hsi_gen.py#L107 I'll need to think a bit more about it.

PR #221, opened 2022-04-12
JCF: Issue #171: put daq_application-specific code in the apps/ subdi…

### LINKED ISSUE #171: See if some headers can be moved out of public include directory
Part of finalizing appfwk's official documentation is making sure that the documentation for all public headers is thorough and up-to-date. In the course of looking into this, I've found a couple of headers which may not, in fact, need to be public:
* `include/Logger.hpp`: it seems like this is an artifact of the days before we had the logging package and can be removed from the package entirely
* `include/Application.hpp` and `include/CommandLineInterpreter.hpp`: these both appear to only be `#include`d by `apps/daq_application.cxx`, in which case we should move them into `apps/`

jcfreeman2:
…rectory + remove obsolete Logger.hpp header

To say a little bit more about "daq_application specific code": basically, the `Application`, `CommandLineInterpreter`, and `DAQModuleManager` classes have all been exclusively used by `daq_application`, not just in `appfwk` but in our overall codebase. What I've done in this PR is move the headers for these classes out of the public header directory and into the `apps/` directory; furthermore these classes are no longer compiled into the public `appfwk` library. 

One quick point: in order for the unit test code to see these headers since they've been moved into `apps/`, I've needed to add lines such as
```
target_include_directories(Application_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
```
since `daq_add_unit_test` currently will include headers from the public `include/` directory, `'src'`, `plugins/`, etc., but not `apps/`. We may want to change this.


eflumerf (approved):
LGTM.
Unit tests pass (except the queue tests which are not failing because of these changes)
Nanorc runs default configuration (daqconf_multiru_gen test)

PR #222, opened 2022-04-13
Use topological sort on the start/stop order between apps

philiprodrigues:
This PR modifies `conf_utils.py` to use the topological sort to set the between-applications start and stop order. It looks like it was mistakenly removed by commit [16994024df21dd02eb6b877ca8de1682672fd54e](https://github.com/DUNE-DAQ/appfwk/commit/16994024df21dd02eb6b877ca8de1682672fd54e)


plasorak (approved):
LGTM

PR #226, opened 2022-04-15
JCF: Issue #171: move the daq_application-specific headers from apps/…

### LINKED ISSUE #171: See if some headers can be moved out of public include directory

jcfreeman2:
… to src/

Based on Giovanna's recommendation. Among other things this gets rid of those inelegant 
`target_include_directories(Application_test PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})`
-style calls in `CMakeLists.txt` which had been needed so that the unit tests could find the headers in the `apps/` subdirectory, since `daq_add_unit_test` already knows to look in `src/`


eflumerf (approved):
LGTM

PR #236, opened 2022-05-31
Update the documentation with some hints on using IOManager.

eflumerf:


jcfreeman2:
Thanks for the update. Just a couple of quick comments:
* In the line starting with "daq_application is designed as a flexible container..." do we want to continue referencing "queues" or replace that with "connections"?
* `register_command`: right now it looks like the only DAQ module which uses the new, third optional argument (the set of valid states) is `RandomDataListGenerator` in listrev, but when people start using that third argument we may want to add mention of this
* "Queues are commonly allocated in init" - modernize this to reference connections?
* A couple of locations in the code examples, it appears that the `qi` variable has been left in where it should have been replaced with `ci`

And _super_ minor details. We can ignore these and still have happy lives, but:
* Doesn't really affect the pedagogy but we could consider linking to a `FakeCardReader.hpp` that's not in the now-obsolete readout repo
* At the bottom, there's the phrase "DAQ Modules" where elsewhere "Modules" isn't capitalized.

jcfreeman2:
LGTM.

jcfreeman2 (approved):

PR #240, opened 2022-07-11
Allow monitoring for all states

mroda88:


mroda88:
Actually, the code written by Giovanna is already state insensitive, so I think that code fixes everything

PR #244, opened 2022-07-20
nlohmann::json::parse the db data

plasorak:


glehmannmiotto (approved):
Please use TLOG_DEBUG for debug messages instead of TLOG.
For the rest the changes look good to me.

PR #246, opened 2022-07-21
Added virtual destructor to quiet compiler warning

bieryAtFnal:
I'm not sure if the compiler warnings that we see with the current code are red herrings, but the changes in this PR quiet them.


eflumerf (approved):
LGTM

PR #250, opened 2022-10-17
Fix the Application_test unit test.

eflumerf:


eflumerf:
I put Wes as a reviewer in case this should be retargeted for a patch/v3.2.x branch 😁

jcfreeman2:
LGTM; I've confirmed that the unit test which had previously been failing works when we switch to this feature branch from `develop`.

PR #251, opened 2022-11-02
Add host in appinfo

mroda88:
in this way we can monitor the business of the cluster using dashboards

### DIFF
--- a/src/detail/Application.hxx
+++ b/src/detail/Application.hxx
@@ -138,8 +138,11 @@ Application::gather_stats(opmonlib::InfoCollector& ci, int level)
   ai.state = get_state();
   ai.busy = m_busy.load();
   ai.error = m_error.load();
-  ai.host = getenv("HOSTNAME");
-
+  auto host = getenv("HOSTNAME");
+  std::string hoststr = "Unknown";
+  if(host) { hoststr = std::string(host); }
+  ai.host = hoststr;
+  
   opmonlib::InfoCollector tmp_ci;
 
   tmp_ci.add(ai);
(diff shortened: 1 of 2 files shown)


src/detail/Application.hxx
```diff
@@ -16,6 +16,8 @@
 
 #include <string>
 
+#include "boost/asio.hpp"
```
eflumerf: Why not `gethostname` or even `getenv("HOSTNAME")`? I'm not sure if including Boost is a good idea here...

mroda88: I can go for `getenv("HOSTNAME")` if we prefer. I used Boost since it's already in the dependencies. But the environmental variable should be a pretty solid solution too.

src/detail/Application.hxx
```diff
@@ -138,7 +138,7 @@ Application::gather_stats(opmonlib::InfoCollector& ci, int level)
   ai.state = get_state();
   ai.busy = m_busy.load();
   ai.error = m_error.load();
-  ai.host = boost::asio::ip::host_name();
+  ai.host = getenv("HOSTNAME");
```
eflumerf: Probably not needed for HOSTNAME, but it may be a good idea to do a not-NULL check and put it into the std::string constructor
```C++
auto host = getenv("HOSTNAME");
std::string hoststr = "Unknown";
if(host) { hoststr = std::string(host); }
ai.host = hoststr;
```

eflumerf (approved):
LGTM

wesketchum (approved):

PR #261, opened 2023-01-23
Add application parameter for connectivity service publish interval, and

eflumerf:
pass it to IOManager. Fix instances in Application_test which were using incorrect app name for given configuration scripts (tension between Application_test and daq_application.sh resolved). Update init to turn off connectivity service for Application test.

Follows https://github.com/DUNE-DAQ/iomanager/pull/35


ron003 (approved):
tested at np04

PR #262, opened 2023-03-14
Remove SIGHUP from signal handler list.

eflumerf:
Commenting out this signal handler appears to resolve many issues with `daq_application` instances hanging around after NanoRC thinks that they should be gone. While not exactly a "solution" for the problem, I think it's a robust enough workaround to merit inclusion for now, as the "real" solution will likely involve lots of work on the CCM side.

### DIFF
--- a/apps/daq_application.cxx
+++ b/apps/daq_application.cxx
@@ -56,7 +56,7 @@ main(int argc, char* argv[])
 
   // Setup signals
   // std::signal(SIGABRT, signal_handler);
-  std::signal(SIGTERM, signal_handler);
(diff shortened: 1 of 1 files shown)


eflumerf:
I have tested with this change on `daq.fnal.gov`, and where it could not reliably complete a single multi-configuration integration test before, it now is capable of running my "run_integration_tests.sh" which runs all of the host-independent integration tests back-to-back.

plasorak:
Ok, after discussion with Alessandro, we agreed we should remove the `SIGTERM` and `SIGHUP` signal handlers.

plasorak (approved):

PR #266, opened 2023-11-07
JCF: given changes due to DUNE-DAQ/daq-cmake#100, make sure that file…

jcfreeman2:
…ConfFacility and dbConfFacility link to cetlib so that all functions in ConfFacility.hpp have definitions

UPSHOT: there's no need to merge this until a version of daq-cmake which includes the PR DUNE-DAQ/daq-cmake#101 has been cut.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,8 +29,8 @@ daq_add_library(DAQModule*.cpp LINK_LIBRARIES ${APPFWK_DEPENDENCIES})
 
 ##############################################################################
 # Plugins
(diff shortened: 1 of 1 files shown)


eflumerf:
Shouldn't the cetlib targets be added to the INTERFACE cmake definition of duneconffacility so that users don't have to worry about it?

eflumerf (approved):
LGTM, depending on which way we resolve https://github.com/DUNE-DAQ/daq-cmake/pull/101

jcfreeman2:
Closing this PR because in https://github.com/DUNE-DAQ/daq-cmake/pull/101 cetlib gets automatically linked in to the plugin if available, thus the change here is no longer necessary.

PR #267, opened 2023-12-11
Make ConfigurationManager and ModuleConfiguration no longer singleton…

eflumerf:
…s. Work on getting Application_test working with TEST schema additions to daq-cmake


eflumerf:
This should not be merged until daq-cmake is tagged with https://github.com/DUNE-DAQ/daq-cmake/pull/110

CMakeLists.txt
```diff
@@ -25,7 +25,10 @@ find_package(appdal REQUIRED)
 set(APPFWK_DEPENDENCIES ${CETLIB} ${CETLIB_EXCEPT} ers::ers logging::logging cmdlib::cmdlib rcif::rcif iomanager::iomanager opmonlib::opmonlib nlohmann_json::nlohmann_json pistache_shared
   coredal::coredal appdal::appdal oksdbinterfaces::oksdbinterfaces okssystem::okssystem)
 
-include_directories(${CMAKE_BINARY_DIR}/coredal ${CMAKE_BINARY_DIR}/appdal)
```
gcrone: Actually, I don't think it should be necessary to have these include_directories with the latest daq-cmake.

include/appfwk/ModuleConfiguration.hpp
```diff
@@ -21,43 +21,39 @@
...
-  };
+class ModuleConfiguration
+{
+  std::vector<const dunedaq::coredal::DaqModule*> m_modules;
+  iomanager::Queues_t m_queues;
+  iomanager::Connections_t m_networkconnections;
+  std::shared_ptr<ConfigurationManager> m_config_mgr;
+
+public:
+  ModuleConfiguration(std::shared_ptr<ConfigurationManager> mgr);
```
gcrone: Should this be explicit? I seem to remember being told that constructors with a single argument should be explicit but I can't remember why.

gcrone (approved):
Looks OK to me

include/appfwk/ModuleConfiguration.hpp
eflumerf: It disables implicit conversions, which is generally a good thing.

CMakeLists.txt
eflumerf: Removed, still builds.

PR #268, opened 2024-01-16
Gcrone/module configuration fix

gcrone:
Fix a couple of problems in ModuleConfiguration. I also removed code related to an earlier schema where DaqModules were Resources and DaqApplications were ResourceSets


eflumerf (approved):
LGTM

PR #269, opened 2024-02-08
Protect from disasters while gather_stats

mroda88:
Add protections against failures in monitoring thread.

### DIFF
--- a/src/detail/DAQModuleManager.hxx
+++ b/src/detail/DAQModuleManager.hxx
@@ -224,21 +224,21 @@ DAQModuleManager::gather_stats(opmonlib::InfoCollector& ci, int level)
   iomanager::NetworkManager::get().gather_stats(ci, level);
 
   for (const auto& [mod_name, mod_ptr] : m_module_map) {
-    opmonlib::InfoCollector tmp_ci;
     try {
+      opmonlib::InfoCollector tmp_ci;
       mod_ptr->get_info(tmp_ci, level);
+      if (!tmp_ci.is_empty()) {
+	ci.add(mod_name, tmp_ci);
+      }
     }
     catch( ers::Issue & i ) {
-      ers::error( FailedInfoGathering(ERS_HERE, mod_name, i) );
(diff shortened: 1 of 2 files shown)


src/detail/DAQModuleManager.hxx
```diff
@@ -225,7 +225,18 @@ DAQModuleManager::gather_stats(opmonlib::InfoCollector& ci, int level)
...
+    catch( ers::Issue & i ) {
+      ers::error( FailedInfoGathering(ERS_HERE, mod_name, i) );
+    }
+    catch( std::exception & ex ) {
+      ers::error( ExceptionWhileInfoGathering(ERS_HERE, mod_name, ex.what()) );
+    }
+    catch( ... ) {
+      ers::error( FailedInfoGathering(ERS_HERE, mod_name) );
+    }
     if (!tmp_ci.is_empty()) {
```
alessandrothea: I suspect that if an exception is thrown `tmp_ci` may not be in an healthy state

alessandrothea: And with that I meant  : in a well know state. Maybe better to move the if statement inside the try catch

src/detail/DAQModuleManager.hxx
```diff
@@ -225,7 +225,18 @@ DAQModuleManager::gather_stats(opmonlib::InfoCollector& ci, int level)
 
   for (const auto& [mod_name, mod_ptr] : m_module_map) {
     opmonlib::InfoCollector tmp_ci;
-    mod_ptr->get_info(tmp_ci, level);
+    try {
+      mod_ptr->get_info(tmp_ci, level);
+    }
+    catch( ers::Issue & i ) {
+      ers::error( FailedInfoGathering(ERS_HERE, mod_name, i) );
```
alessandrothea: Error or warning? I would go with warning maybe

alessandrothea (changes requested):

alessandrothea (approved):

bieryAtFnal (approved):
LGTM2 (regression testing)

PR #288, opened 2024-08-05
Feature/opmon

mroda88:


glehmannmiotto (approved):
ready to go into develop, in my opinion. Once the Run Control Interface will have been re-implemented, that class should become the monitorable object holding the state.

PR #291, opened 2024-08-26
Add more detail on the motivation for Action Plans, and note that the

eflumerf:
FSMCommand objects should be treated as external inputs.


glehmannmiotto (approved):
LGTM

PR #295, opened 2024-09-30
Remove the `--name` and `--session` from the cli, and use env instead

plasorak:


eflumerf (approved):

apps/daq_application.cxx
```diff
@@ -79,19 +79,34 @@ main(int argc, char* argv[])
...
+  std::string session_name = "";
+
+  char* app_name_c = getenv("DUNEDAQ_APPLICATION_NAME");
+  char* session_name_c = getenv("DUNEDAQ_SESSION");
+
+  if (!app_name_c || strcmp(app_name_c, app_name.c_str()) == 0) {
+    ers::error(appfwk::EnvironmentVariableNotFound(ERS_HERE, "DUNEDAQ_APPLICATION_NAME"));
+    exit(1);
+  }
+  if (!session_name_c || strcmp(session_name_c, session_name.c_str()) == 0) {
```
alessandrothea: Would it be possible to check for the existence of both variables first and then throw the error/exit?

apps/daq_application.cxx
```diff
@@ -79,19 +79,34 @@ main(int argc, char* argv[])
...
-  setenv("DUNEDAQ_APPLICATION_NAME", args.app_name.c_str(), 0);
-  setenv("DUNEDAQ_PARTITION", args.session_name.c_str(), 0);
+  // Get the application and session name from the environment.
+  std::string app_name = "";
+  std::string session_name = "";
+
+  char* app_name_c = getenv("DUNEDAQ_APPLICATION_NAME");
+  char* session_name_c = getenv("DUNEDAQ_SESSION");
+
+  if (!app_name_c || strcmp(app_name_c, app_name.c_str()) == 0) {
```
alessandrothea: What is this check for?
```
strcmp(app_name_c, app_name.c_str()) == 0
```
To ensure that the variable is not empty?

alessandrothea (changes requested):

apps/daq_application.cxx
plasorak: Yes, sorry, my C++ is a bit rusty

alessandrothea (approved):

PR #298, opened 2024-10-04
Dont ignore the `use_connectivity_service` flag

plasorak:
Correct a little typo (with Eric's help) that made the app always use the connectivity server, regardless of `session.use_connectivity_server`.


eflumerf (approved):
Builds, unit tests pass

PR #300, opened 2024-10-08
Revert "Remove the `--name` and `--session` from the cli, and use env instead"

plasorak:
Reverts DUNE-DAQ/appfwk#295 requires: https://github.com/DUNE-DAQ/confmodel/pull/47


eflumerf (approved):
Builds, unit tests pass, and once I restored the iomanager::shutdown() call, regression tests pass.

PR #307, opened 2024-11-12
Application_test fixes

eflumerf:

PR #308, opened 2024-11-12
Repeat of #307, on v5.2.1 patch branch

eflumerf:


eflumerf:
Needs the change from https://github.com/DUNE-DAQ/opmonlib/pull/43 to resolve segfaults in one of the test cases.

jcfreeman2:
Unit test now succeeds getting the session. Having said that, there's a new error, but I'll file a separate Issue for that. Approved.

jcfreeman2 (approved):

PR #314, opened 2025-01-09
Add try catch block

### LINKED ISSUE #312: Application::init is used unprotected
Application::init() is called without any try/catch protection https://github.com/DUNE-DAQ/appfwk/blob/6ce3c7fa06f19812670cd1ab63b86fd825afff4f/apps/daq_application.cxx#L89 

Unless a protection happens at module specific level, there is not way to prevent the application to crash.

mroda88:
The creation of an application was not protected against issues and therefore things were not reported correctly in case of fatal error. 
This is a simple solution, but it works.


eflumerf (approved):
LGTM, unit tests pass

PR #316, opened 2025-02-13
Patch/fddaq v5.2.x

andrewmogan:


jcfreeman2 (approved):
Merging with `fddaq-v5.2.1` having been released today. Also note that `git` considers there to be a conflict in the merge with `src/ConfigurationManager.cpp`, but the right course of action is to just take the already-existing `src/ConfigurationManager.cpp` on the `develop` branch.

PR #317, opened 2025-02-18
Introduce Configurable base class for Application

eflumerf:
Open question: Do we move Configurable to its own header and make more `appfwk` classes inherit from it? Candidates include `DAQModuleManager` and `DAQModule`. Note that DAQModuleManager_test re-initializes DAQModuleManager with different ConfigurationManager instances, however this could probably be changed to use different DAQModuleManager instances...

### DIFF
--- a/src/Application.cpp
+++ b/src/Application.cpp
@@ -31,7 +31,7 @@ Application::Application(std::string appname,
                          std::string cmdlibimpl,
                          std::string confimpl)
   : Configurable(confimpl, appname, session) 
-  , OpMonManager(session, appname, m_config_mgr->session()->get_opmon_uri()->get_URI(appname))
+  , OpMonManager(session, appname, get_config_manager()->session()->get_opmon_uri()->get_URI(appname))
   , NamedObject(appname)
   , m_state("NONE")
   , m_busy(false)
@@ -43,12 +43,10 @@ Application::Application(std::string appname,
   m_runinfo.set_run_time(0);
 
   m_cmd_fac = cmdlib::make_command_facility(
-    cmdlibimpl,
-    session,
-    m_config_mgr->session()->get_connectivity_service()
+    cmdlibimpl, session, get_config_manager()->session()->get_connectivity_service()
   );
 
-  set_opmon_conf(m_config_mgr->application()->get_opmon_conf());
+  set_opmon_conf(get_config_manager()->application()->get_opmon_conf());
 
   TLOG() << "confimpl=<" << confimpl << ">\n";
 }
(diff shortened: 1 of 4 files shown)


mroda88:
technically, exposing a protected member is against our styleguide, if I remember correctly. Do we need a getter function? But this looks fine to me. But let's check this small detail first.

jcfreeman2:
tl;dr : Great catch @mroda88 . It's not the end of the world if `m_config_mgr` were exposed to derived classes as `protected`, but it might be nice to wrap it in a simple getter function.

The Style Guide does indeed say that data should always be `private`. In the interests of keeping that document as short as possible I left out explanation, but basically the ideas behind this rule are (1) that `protected` data "might as well" be `public`, since you can always derive a class and access the `protected` data that way, and (2) you can't change the representation of the data like you could with an encapsulating getter function. A bit more can be found in the Style Guide's ancestor document, the Google Style Guide,  (https://google.github.io/styleguide/cppguide.html#Access_Control)

Now. (1) _really_ only applies if, e.g., there were a security reason to keep a variable invisible to outside users of your class. We're all friends here, so I don't think that's the case. (2) would apply if, e.g., you wanted to be able to decide later on that you could use a `unique_ptr` instread of a `shared_ptr` to the configuration manager, and knew that users of your class wouldn't and shouldn't care about this implementation detail. No idea if that applies here, honestly. If it does, and no user of the class will ever want to own the configuration manager, then this is the one time it's ok to use a (`const`) raw pointer.

mroda88:
Given that we are dealing with transitions I certainly welcome the safety of a full protection here, i.e. private data member and public const getter function.

alessandrothea:
The `ConfigurationManager` object that `Configurable` for internal consumption, let's not get ahead of ourselves.
@mroda88 I have no idea what you mean with "transition" or "full protection" in this context.

alessandrothea:
> Open question: Do we move Configurable to its own header and make more `appfwk` classes inherit from it? Candidates include `DAQModuleManager` and `DAQModule`. Note that DAQModuleManager_test re-initializes DAQModuleManager with different ConfigurationManager instances, however this could probably be changed to use different DAQModuleManager instances...

@eflumerf `DAQModuleManager` constructor takes a `std::shared_ptr<ConfigurationManager>` argument while this new `Configurable` class is an artefact to enable early construction of a `ConfigurationManager` for the consumption of other classes in the inheritance chain. I don't imagine we want `DAQModuleManager` to build it's own copy of a `ConfigurationManager` or have multiple `ConfigurationManager` objects at all, do we?

alessandrothea (approved):
LGTM

Just a couple of comments:
- `Configurable`: the name suggests the existence of a `configure(...)`  method of sort, so it doesn't feel quite right. I don't have better suggestions so I can't really complain :)
- As I said on slack, this seems the most elegant solution given the way `Application` relies on inheritance. That said, having a quadruple (?) inheritance feels like a lot and it may not scale much further...

PR #321, opened 2025-02-27
Catch DAQModule init errors in DAQModuleManager, add the name of the

### LINKED ISSUE #312: Application::init is used unprotected

### LINKED ISSUE #320: No indication of what happened when TCMaker throws an exception
The steps listed below recreate a problem in which the `tc-maker-1` process dies because there is a missing configuration parameter.  The reason that I'm filing this Issue is that, when this happens, there is nothing in the application log files that indicate that the TriggerCandidateMaker threw an exception that was not handled ([here](https://github.com/DUNE-DAQ/triggeralgs/blob/33312c5d25ed494923695953a2823d9a82ed3c05/include/triggeralgs/TriggerCandidateMaker.hpp#L118)).  When I do a test run, the `tc-maker-1` process just silently goes away.  Why isn't there more of an indication of what happened?

Maybe I answered my own question when I said "_unhandled_ exception".  If the application framework has no way to indicate what happened in this case, I'll add an `ers::error` call immediately before the TCM code throws the exception so that in the future, we'll have at least a little indication of what happened.

```
DATE_PREFIX=`date '+%d%b'`
TIME_SUFFIX=`date '+%H%M'`

source /cvmfs/dunedaq.opensciencegrid.org/setup_dunedaq.sh
setup_dbt latest_v5
dbt-create -n NFD_DEV_250226_A9 ${DATE_PREFIX}FDDevTest_${TIME_SUFFIX}
cd ${DATE_PREFIX}FDDevTest_${TIME_SUFFIX}/sourcecode

git clone https://github.com/DUNE-DAQ/appmodel.git -b develop
git clone https://github.com/DUNE-DAQ/daqsystemtest.git -b develop
git clone https://github.com/DUNE-DAQ/dfmodules.git -b develop
git clone https://github.com/DUNE-DAQ/trigger.git -b develop
git clone https://github.com/DUNE-DAQ/triggeralgs.git -b kbiery/tcmaker_avoid_crash
cd ..

sed -i 's/.*The output TC type name.*//' sourcecode/appmodel/schema/appmodel/trigger.schema.xml 

dbt-workarea-env
dbt-build -j 12
dbt-workarea-env

daqconf_set_connectivity_service_port local-1x1-config config/daqsystemtest/example-configs.data.xml

mkdir rundir
cd rundir

drunc-unified-shell ssh-standalone config/daqsystemtest/example-configs.data.xml local-1x1-config boot --no-override-logs wait 5 conf wait 3 start --run-number 101 enable-triggers wait 10 disable-triggers drain-dataflow stop-trigger-sources stop scrap terminate

egrep -i 'error|warn' log*.txt
```

eflumerf:
failing module. Add new Issue type to DAQModule.hpp to carry this information.

Move Application constructor outside of try..catch block in daq_application, apparently the destructor of Application was interfering with the ers::fatal report of the issue which caused its destruction.

### DIFF
--- a/apps/daq_application.cxx
+++ b/apps/daq_application.cxx
@@ -81,15 +81,15 @@ main(int argc, char* argv[])
 
   // from now on, it's possible to use ERS messages
 
+  // Create the Application
+  appfwk::Application app(
+    args.app_name, args.session_name, args.command_facility_plugin_name, args.conf_service_plugin_name);
+
   try {
-    // Create the Application
-    appfwk::Application app(args.app_name, args.session_name,
-			    args.command_facility_plugin_name,
-			    args.conf_service_plugin_name);
     
     app.init();
     app.run(run_marker);
-  } catch ( const ers::Issue & e ) {
+  } catch ( ers::Issue & e ) {
     ers::fatal(appfwk::ApplicationFailure(ERS_HERE, args.session_name, args.app_name, e));
   }
 
--- a/include/appfwk/DAQModule.hpp
+++ b/include/appfwk/DAQModule.hpp
@@ -21,9 +21,9 @@
 
 #include "appfwk/ConfigurationManager.hpp"
 
-#include "utilities/NamedObject.hpp"
-#include "opmonlib/MonitorableObject.hpp"
 #include "logging/Logging.hpp" // NOTE: if ISSUES ARE DECLARED BEFORE include logging/Logging.hpp, TLOG_DEBUG<<issue wont work.
+#include "opmonlib/MonitorableObject.hpp"
+#include "utilities/NamedObject.hpp"
 
 #include "cetlib/BasicPluginFactory.h"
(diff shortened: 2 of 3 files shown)


mroda88 (approved):
Oh, wow that was nasty. 
Good catch!

bieryAtFnal (approved):
LGTM.  I verified that with these changes, I see the expected exception message in the tc-maker-1 log file.  I also verified that when I remove my temporary modification that causes the TriggerCandidateMaker to fail, I see all of the integtests in the daqsystemtest repo run cleanly.  Thanks!

Issue #324, opened 2025-03-08
Clarification about Application runinfo opmon data


alessandrothea:
`Application` published the `dunedaq::rcif::opmon::RunInfo` from run control interface, that includes information like `running` state, `run_number` and `run_time`. 
Since the `Application` code doesn't have any concept of "run" it doesn't seem very appropriate for this information to be published here.
Also, why does the information focusses on the "running" state, instead of the present state and time spent in this state?
Why is the run number a relevant information to publish at `Application` level?

PR #326, opened 2025-04-03
Implement review suggestions in #322

### LINKED ISSUE #322: General code review

eflumerf:


eflumerf:
Note that Issues #324, #327, #328, and #329 cover items that are not implemented here so that the changes remain solely in `appfwk`

jcfreeman2:
These changes look great. I have just a couple of comments, roughly in decreasing order of importance:
* When I stripped out unneeded code via commit `13cd18ad7` on this branch last month, I accidentally left `ConfFacility.hpp` in there; it should be removed. 
* I see there are still interspersed comments in the ERS issue; was this an oversight, or is there a reason they're still in there?
* While I like that the body of `make_module` was moved to `DAQModule.hxx`, in the spirit of the `*.hpp` file presenting the interface for users, could we also have a declaration of the `make_module` signature in `DAQModule.hpp` right below the `DAQModule` class declaration?
* We can now drop the include of `nlohmann/json.hpp` from `daq_application.cxx`
* Some of the new comments at the top of headers (e.g., in `DAQModule.hpp`) have lines sufficiently long that line wrapping becomes an issue

eflumerf:
That last commit message is a lie (stupid scripts...) I fixed the issues John mentioned, except for the comments which were re-formatted using clang-format which apparently has a ColumnWidth of 120

jcfreeman2:
Thanks, approved.

jcfreeman2 (approved):

PR #332, opened 2025-05-27
Simplified genmodules signature

alessandrothea:
This PR removes the `confdb` and `dbfile` arguments of the `SmartDaqApplication::generate_modules(...)`.


mroda88 (approved):
Tested with runs at EHN, all the integration tests, all the unittests. All work. EHN1 configuration generation also worked.

PR #333, opened 2025-05-29
Increment version number. Mid number because this package is reacting…

eflumerf:
… to an API change in appmodel


andrewmogan (approved):
LGTM

PR #335, opened 2025-06-04
Johnfreeman/new daq cmake for v5.3.x

jcfreeman2:
The following PR enables this package to build on its patch branch against `daq-cmake` `v3.1.0`, which contains changes described in DUNE-DAQ/daq-cmake#155. 

This is only to be merged by Software Coordination.


andrewmogan (approved):
I successfully built an area based on `NFDT_DEV_250604_A9` and ran smoke tests without issues.

PR #337, opened 2025-06-10
Remove the typedef data_t = nlohmann::json from DAQModule.

### LINKED ISSUE #336: It would be helpful if DAQModule::data_t had a more unique and descriptive name
Searching our code for "data_t" returns matches to "data_type", "data_table", "update_metadata_to_bookkeeper", and possibly other strings, in addition to bare "data_t" strings.

It would be helpful if the name of this alias were more unique and search-able.  And, possibly more descriptive.

eflumerf:
Note that this changes the DAQModule API and results in related (simple, automatable) changes in many other repos.

Resolves #336


eflumerf:
For automatically updating, I checked out all of DUNE-DAQ into a working area, and ran this:
```
sed -i 's/data_t\&/nlohmann::json\&/g' `find . -type f -name *.*pp|grep -l 'data_t&'`
```

jcfreeman2:
It seems this is solving a different problem than the one Kurt posed? If we want to make `data_t` more _searchable_ then just dropping the `typedef` and having the underlying `nlohmann::json` appear in its place wouldn't do the trick; perhaps renaming `data_t` to `daqmodule_data_t` or `daqmodule_config_t` would?

eflumerf:
To be discussed and scheduled at SW I&T since it does affect many repos.

jcfreeman2:
Also, since in our naming conventions we have "The names of all types — classes, structs, type aliases, enums, and type template parameters — use Pascal case." let's revise my suggestions to `DAQModuleData_t` and `DAQModuleConfig_t`.

eflumerf:
Closing because I agree with John that this should really just be a better-named typedef.

PR #338, opened 2025-06-11
Update DAQModule_test and return it to the build

### LINKED ISSUE #329: Reinstate appfwk tests
Currently, DAQModule_test is not built and daq_application_test.sh is non-functional. We should investigate whether these tests are still necessary, and, if so, update them to work with the current system. Part of #322.

eflumerf:
Resolves #329


jcfreeman2 (approved):
`DAQModule_test`s all pass; approved.

PR #342, opened 2025-07-15
ActionPlan validation improvements

### LINKED ISSUE #341: Remove the word "Error" from ERS Issue declaration
https://github.com/DUNE-DAQ/appfwk/blob/1f8ce77ccdf23b8e068b8096218cfb26dd87d80e/include/appfwk/ConfigurationManager.hpp#L38C1-L41C81

eflumerf:
Re-word ERS Issue to remove problematic word "Error"
Promote "Module does not exist" when no matching modules of class are in the application from info to warning.
Add cross-check for whether all modules that have a command are correctly referenced by the ActionPlan, if defined for that command.

Resolves #341


eflumerf:
Includes changes from #344

mroda88 (approved):
LGTM, it was also tested with a dedicated nightly

PR #343, opened 2025-07-15
remove spurious include of confmodel/ResourceBase.hpp

jcfreeman2:


mroda88 (approved):
Tested with NP02 configurations

PR #344, opened 2025-07-16
Add more examples on ActionPlan execution

eflumerf:
Highlight the difference between the two execution modes, and note that the order within steps for dispatching commands is dependent on the order modules are declared to the application.


bieryAtFnal:
Looking at the full ActionPlan.md document...

[After writing the following notes, and running out of steam at some point, I realized that I should offer to make a branch with proposed changes for consideration.  I can to that, if it would be helpful.]

1. In the "Defining an ActionPlan" section, I don't understand the use of the word "parallel" in the schema definition of an ActionPlan.  The steps are always executed in series.  Step1 happens before Step2, etc.  I'm confused why this description doesn't say something like "A set of steps that specify the order in which an application delivers a command to its modules".  (or "executes a command on its modules")
2. In the same section of the document, I'm not sure why the description of an `execution_policy` doesn't say "How the application should deliver the command to the modules within each step of the action plan".
3. In the last paragraph in that section, we are told that "SmartDaqApplications are only allowed to use ActionPlans which group modules by type", but we are given no information about which of our common DAQ applications fall into this category.  A quick (non-exhaustive) look through `appmodel/schema/appmodel/application.schema.xml` seems to imply that all of the common DAQ applications are SmartDAQApplications.  If that is true, shouldn't the text in the document say that?
4. In the "Example test/config/appfwk.data.xml" section, there is a sentence that describes the FSMCommand entry in the example.  It would be helpful to have sentences that describe the "ActionPlan" and "DaqModulesGroupByType" entries, also.  In addition, it would be very helpful to have some indication of which text in the example is boiler-plate and which is specified by the user.  That is, what parts of the example would need to be copied exactly as show, and which could be changed by the user if they wanted to create their own.
5. In the "The ActionPlans are associated with the Application instance as follows:" example, it would be helpful to have some sort of highlighting to show that the entries in the "action_plans" are what the reader should focus on.
6. I'm confused by the Note that "DAQModules register their action methods in the same way as before, however the specification of valid states for an action has been removed".  For a new user who doesn't have knowledge of how things were done before, what does this statement tell us?
7. I would find it very helpful to change "For each step of an ActionPlan, whether executing in series or in parallel," to "Within each step of an ActionPlan, whether executing in series or in parallel,".
8. I'm very confused by the text after the first Further Example, "This ActionPlan consists of four steps, with each step being executed in parallel."  I would expect this to say "This ActionPlan consists of four steps that are executed in series.  Within each step, the command is executed in parallel for the modules that are included in that step."

bieryAtFnal (approved):
Looks good to me.  Thanks for the changes.

PR #346, opened 2025-07-22
Add processing for optional DaqModulesGroup

eflumerf:
Follows https://github.com/DUNE-DAQ/confmodel/pull/76, implements "this DAQModule is optional" behavior requested for TPG use.


eflumerf:
Includes changes in #342

mroda88 (approved):
Tested with EHN1 and the changes are not disruptive for the run. Although we need to rework some of the configurations to specify action plans specific for each applications.

PR #352, opened 2025-09-30
Improve ConfigurationManager and DAQModuleManager API for ActionPlan validation

### LINKED ISSUE #349: Think about refactoring ActionPlan validation
Right now, ActionPlan validation is part of the ConfigurationManager/DAQModuleManager construction process, and is duplicated in `check_action_plans`. It would be much better if the validation were to occur in a dedicated method of ConfigurationManager or DAQModuleManager so that it can be maintained in a single location, so that the duplication does not result in differing functionality.

eflumerf:
Validation can be run without initializing DAQModules. Create ValidationReports in addition to ers Issues, so that check_action_plans can use DAQModuleManager API to validate ActionPlans without reimplementing validation logic.

Resolves #349


mroda88 (approved):
LGTM

bieryAtFnal (approved):
I haven't looked at the code changes, but I've verified that `check_action_plans` continues to work and regression tests continue to pass.

PR #353, opened 2025-10-07
SmartDaqApplication inheritance

gcrone:
Simplify `ConfigurationManager` with `SmartDaqApplication` only differeing from a non smart `DaqApplication` by a call to `generate_modules`.

Requires https://github.com/DUNE-DAQ/appmodel/pull/248


alessandrothea (approved):
LGTM

